PR #4, opened 2014-03-06
[#LIB-158] added audit answer to audit logging

mjuecker:
Instead of returning just nothing and printing a stack trace to System.err I'd like to return an object describing the status of the audit, including the audited message and an exception that possibly occurred while building or sending the message.


hczedik:
Closed. The code has been changed a lot in the meantime. Please open a new updated pull request should that change still be relevant.

PR #5, opened 2014-03-16
Adding in a Sigmoid Lookup

plooney:
Dear Gunter,

I have been trying to view images with SIGMOID lookup tables using dcm4che. I added a task on the JIRA for this. 

http://www.dcm4che.org/jira/browse/LIB-159

I propose this as a solution.


hczedik:
Needs to be reviewed in the context of http://www.dcm4che.org/jira/browse/LIB-159
(Currently not planned in.)

PR #7, opened 2014-04-22
Change package to match class in repo

jed204:
Packages were updated to dcm4che3, the rle reader was not being loaded since the code cannot find the class.


hczedik:
Has already been fixed in the meantime.

PR #18, opened 2015-06-12
Proxy Manager for Connection

maurrysonn:
**Goal :** delegate "Proxy Handshake" of *Connection* class in a simple manager (strategy pattern) and get "entry point" to define new handshake algorithm.

In summary, I created a new interface *ProxyManager* with the *doProxyHandshake(...)* method and defined a *CommonProxyManager* class with the current implementation of *Connection* class.
We can now define new implementation (with new manager) and programmatically set the new algorithm to realize a custom handshake.
So, I don't change the current mechanism and just create entry point.

I need this to implement a Digest authentication (and maybe more).
What do you think about ?

### DIFF
--- a/dcm4che-net/src/main/java/org/dcm4che3/net/Connection.java
+++ b/dcm4che-net/src/main/java/org/dcm4che3/net/Connection.java
@@ -63,8 +63,8 @@ import org.dcm4che3.conf.core.api.ConfigurableClass;
 import org.dcm4che3.conf.core.api.ConfigurableProperty;
 import org.dcm4che3.conf.core.api.ConfigurableProperty.Tag;
 import org.dcm4che3.conf.core.api.LDAP;
-import org.dcm4che3.net.proxy.BasicProxyManager;
 import org.dcm4che3.net.proxy.ProxyManager;
+import org.dcm4che3.net.proxy.ProxyService;
 import org.dcm4che3.util.SafeClose;
 import org.dcm4che3.util.StringUtils;
 import org.slf4j.Logger;
@@ -221,7 +221,8 @@ public class Connection implements Serializable {
     )
     private Protocol protocol = Protocol.DICOM;
 
-    private ProxyManager proxyManager = new BasicProxyManager();
+    // TODO AP - CONFIG PROXY FOR CONNECTION
+    private ProxyManager proxyManager = ProxyService.getInstance().getDefaultProxyManager();
     
 	private static final EnumMap<Protocol, TCPProtocolHandler> tcpHandlers =
             new EnumMap<Protocol, TCPProtocolHandler>(Protocol.class);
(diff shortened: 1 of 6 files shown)


Hesham-Elbadawi:
Hi Amaury,

I reviewed the code and can't help but think that despite this is a good point that you have with the possibility to change implementation for handshake to allow for more than basic HTTP proxy authentication your implementation only allows for CommonProxy Manager to be always utilized.
To generalize the implementation to allow for loading different implementations, i would take a look at how the imageio implementation in the dcm4che-imageio looks like.

It uses Service loader to load different implementations of a specific interface.
Such implementations can be provided inside any jar which contains the Manifest.services folder with interfacenamed file under it that contains the implementation class path inside

Please use this example to be able to load different implementations and put the common one as default.
Then once you have the implementations using service.load method you can iterate over them and select which proxymanager to call depending on some authentication criteria available to one and not the other.

please update your implementation as discussed then we can merge it.

maurrysonn:
Hi Hesham,

Thanks for your time and your review.

After your response, I thought about use the ServiceLoader but one question remains : how can you choose (and set) the implementation that you would to use for a connection ?

In the case of ServiceLoader, we can't set a specific implementation unless if you load a unique implementation. So we need several fields in *Connection* class for configuration API (i.e *dicomHTTPProxyProviderName* and *dicomHTTPProxyProviderVersion*) and include registration informations in *ProxyService*.

In my solution, you just create an instance of *ProxyManager* and you set programatically the manager with the *setProxyManager()* method of *Connection* class. By default, it use the *CommonProxyManager* implementation (define during initialization). We can also imagine configuration fields but it's not required.

So, you think ServiceLoader is really best solution ?
If yes, I do it now... :-)

Hesham-Elbadawi:
Hi Amaury,

I thought we were talking about a service loader for ProxyManager
implementation and not connection implementation.

you don't need to change the implementation for the connection at all. All
you have to do is load implementations for the ProxyManager via the service
loader and iterate over them in the connection class same as you did when
you created the CommonProxyManager in the Connection class but instead you
would have to select your implementation you loaded.

you still set programmatically but from a loaded collection of ProxyManager
implementations.

How to select gets tricky because then you would need to either check the
remote proxy authentication protocol and then choose implementation
appropriate to that or you can create a configuration parameter on the
connection to choose always one configurable implementation which is still
loaded through service loader.

I don't think service loader is the best solution because if you use cdi
for example you could just create a producer for the right implementation
but the service loader approach is the java standard before cdi.


Kind Regards,
Hesham

maurrysonn:
Hi Hesham,

I have completed modifications.
So, I added ServiceLoader for loading *ProxyManager* implementations and updated SPI.
Selection of implementation is done in *Connection* class by either the setter method or by the 2 configuration fields (httpProxyProviderName and httpProxyProviderVersion).

For information, the version is not required and, in this case, the last version of the provider is selected. Moreover, the default implementation (corresponding to the current implementation of Basic auth) was renamed to BasicProxyManager.

I remain available for exchange,
Kind Regards,
Amaury

Hesham-Elbadawi:
Hi Amaury,

Will revise the implementation and get back to you as soon as possible.

Kind Regards,
Hesham

PR #22, opened 2015-08-25
[LIB-375] Tags "Photometric Interpretation" and "Planar Configuration" are not passed to decoder

hczedik:
Pass planar configuration (banded) to decompressor via image type.

Review request for Gunter.


hczedik:
The fix is incorrect. The data is actually corrupted (see http://dcm4che.org/jira/browse/LIB-375?focusedCommentId=18722&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-18722 ) -> Closed

PR #25, opened 2015-10-19
Fixes close stream in finally.

nroduit:


umberto-cappellini:
will slightly rearrange this, but it's a good point, thanks

PR #27, opened 2015-11-29
Handle a list of readers/writers for a tsuid or a mime type

nroduit:
Returns a list filtered according to the system and the processor architecture.


hczedik:
On hold for now. Let's make the the new codecs (e.g. org.weasis.jpeg.NativeJLSImageReader) available first!

PR #30, opened 2016-04-21
Added additional logging when an Exception occurs to aid in …

KevinDougan:
…determining the root cause of the problem.

I found that the existing code was not displaying the error message and was swallowing the Exception without displaying the Stack Trace, so I had to debug the code to figure out what was wrong.

### DIFF
--- a/dcm4che-tool/dcm4che-tool-stowrs/src/main/java/org/dcm4che3/tool/stowrs/StowRS.java
+++ b/dcm4che-tool/dcm4che-tool-stowrs/src/main/java/org/dcm4che3/tool/stowrs/StowRS.java
@@ -200,8 +200,7 @@ public class StowRS {
             try {
                 metadata = parseJSON(file.getPath());
             } catch (Exception e) {
-                LOG.error("error parsing metadata JSON file {}: {}", file, e.getMessage());
-                e.printStackTrace();
+                LOG.error("error parsing metadata JSON file {}: {}", file, e);
                 return;
             }
         } else if (mediaType == StowMetaDataType.XML) {
@@ -214,8 +213,7 @@ public class StowRS {
                 if (fmi != null)
                 	metadata.addAll(fmi);
             } catch (Exception e) {
-                LOG.error("error parsing metadata XML file {}: {}", file, e.getMessage());
-                e.printStackTrace();
+                LOG.error("error parsing metadata XML file {}: {}", file, e);
                 return;
             }
         } else {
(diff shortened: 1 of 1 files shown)


dcm4che-tool/dcm4che-tool-stowrs/src/main/java/org/dcm4che3/tool/stowrs/StowRS.java
```diff
@@ -200,7 +200,8 @@ private void stowMetaDataAndBulkData(File file) {
             try {
                 metadata = parseJSON(file.getPath());
             } catch (Exception e) {
-                LOG.error("error parsing metadata JSON file {}", file);
+                LOG.error("error parsing metadata JSON file {}: {}", file, e.getMessage());
+                e.printStackTrace();
```
hczedik: Hi, Kevin!
As we are already using a Logger here, I wouldn't use e.printStackTrace() (which will print to standard out).
I believe if you do it like this the stack trace should also be visible:
`LOG.error("error parsing metadata JSON file {}: {}", file, e);`
Can you try that out please.

hczedik:
Please do not merge yet, see my inline-comment discussion first.

KevinDougan:
Your suggestion has been added.

dcm4che-tool/dcm4che-tool-stowrs/src/main/java/org/dcm4che3/tool/stowrs/StowRS.java
```diff
@@ -200,7 +200,7 @@ private void stowMetaDataAndBulkData(File file) {
             try {
                 metadata = parseJSON(file.getPath());
             } catch (Exception e) {
-                LOG.error("error parsing metadata JSON file {}", file);
+                LOG.error("error parsing metadata JSON file {}: {}", file, e);
```
hczedik: For slf4j loggers will display the full exception, including stacktrace, even if you do not add the placeholder to the message itself:
`LOG.error("error parsing metadata JSON file {}", file, e);`
Please give that a try.

KevinDougan:
Hi there,

When I use your latest suggested format, it does not print the Stack Trace, but only prints the message. I think it's more valuable to have the Stack Trace there, especially since this is an Error situation, so I think it would be good to go with the currently committed format. Let me know if you agree.

Thanks,
Kevin

KevinDougan:
OK, I just pushed another change up, so it should be good to go now.

Thanks,
Kevin

PR #32, opened 2016-08-17
[LIB-433] Use of multiple character sets may lead to incorrectly en- and decoded strings

mbaldauf:
Proposed solution to fix [[LIB-433]](http://www.dcm4che.org/jira/browse/LIB-433)

I revised and extended the org.dcm4che3.data.SpecificCharacterSet class to:
- provide support for code extension techniques conforming to DICOM 2016c (G0 and G1 code elements may come from different character sets)
- fix the switched esc sequences of ISO 2022 IR 87 and ISO 2022 IR 159 (should be G0 esc sequences)
- fix the defined term of the default character repertoire (no value required; ISO_IR 100 should not be used, as the default character repertoire is ISO_IR 6)
- en- and decode unsupported characters according to the DICOM standard (DICOM PS3.5 2016c, 6.1.2.3 Encoding of Character Repertoires)
- encapsulate the invocation of the default/first character set for delimiters and control characters (e.g. no need to pass delimiters to the encode() method)
- avoid the inclusion of an escape sequence at the end of a data element
- add the TAB character as control character (added in DICOM 2015b)
- handle two special cases regarding ISO 2022 IR 13 (G0 identical to ISO 2022 IR 6 except two characters)

Furthermore, I adapted the org.dcm4che3.data.StringValueType class to remove the use of delimiters (the en- and decoder methods only needs to know if the VR is PN).
Last but not least, the org.dcm4che3.data.SpecificCharacterSetTest class was adapted and extended.

### DIFF
--- a/dcm4che-core/src/main/java/org/dcm4che3/data/SpecificCharacterSet.java
+++ b/dcm4che-core/src/main/java/org/dcm4che3/data/SpecificCharacterSet.java
@@ -77,16 +77,9 @@
 package org.dcm4che3.data;
 
 import java.io.UnsupportedEncodingException;
-import java.lang.ref.SoftReference;
 import java.nio.ByteBuffer;
-import java.nio.CharBuffer;
-import java.nio.charset.CharacterCodingException;
 import java.nio.charset.Charset;
-import java.nio.charset.CharsetEncoder;
-import java.nio.charset.CoderResult;
-import java.nio.charset.CodingErrorAction;
 import java.util.Arrays;
-import java.util.StringTokenizer;
 
 /**
  * @author Gunter Zeilinger <gunterze@gmail.com>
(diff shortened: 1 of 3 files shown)


hczedik:
Hi Matthias,
thanks for your pull request. We plan to look into it. In the meantime some small changes have already happened on SpecificCharacterSet and your changes do not apply cleanly any more.
Could you rebase your changes?

mbaldauf:
Hi Hermann,
Thanks for your feedback. I rebased my changes and, as I am new to Git, hope that everything worked out well. If you have any questions or need further information, please feel free to contact me.

gunterze:
Your proposed solution decreases performance significantly (e.g. testEncodeKoreanLongText() tooks about 20 times longer!) and the encoding results differs from the canonical results of examples given in DICOM Part 5 (e.g. http://dicom.nema.org/medical/dicom/current/output/chtml/part05/sect_I.3.html ).

mbaldauf:
Hello Gunter,
First of all, thanks for having a look at the proposed solution, giving feedback, and incorporating many of my suggestions into your changes, which I really appreciate.

I replied/commented on the points you raised in the corresponding JIRA ticket [[LIB-433]](http://www.dcm4che.org/jira/browse/LIB-433). In my opinion there are still some issues to clarify regarding the current implementation. Therefore, I would be pleased if you could have a look at it. Furthermore, I would like to point out that I revised my originally proposed solution in order to fix and improve some aspects regarding en- and decoding.

King regards,
Matthias

PR #34, opened 2016-08-25
[LIB-435] wrong mapping in StandardElementDictionary.tmTagOf()

daniel-h-rech:
Fix for [http://www.dcm4che.org/jira/browse/LIB-435](http://www.dcm4che.org/jira/browse/LIB-435)


hczedik:
Thank you!

PR #35, opened 2016-10-28
Content-type needs to be properly quoted for StowRS request

davegwatson:
Per RFC 2387 section 3.4:
(https://tools.ietf.org/html/rfc2387#section-3.4)

   Note that the parameter values will usually require quoting.  Msg-id
   contains the special characters "<", ">", "@", and perhaps other
   special characters.  If msg-id contains quoted-strings, those quote
   marks must be escaped.  Similarly, the type parameter contains the
   special character "/".


hczedik:
Thank you David!

PR #36, opened 2016-11-02
Proposal: add support for specifying a stop tag matching strategy …

mike-mcfarland:
… when reading DICOM datasets via DicomInputStream. This would allow clients to perform [in/ex]clusive reads up to specific tags, without prior knowledge of which stop tags are actually available in the underlying stream (e.g. when using the MatchOrExceed strategy, dis.readDataset( -1, Tag.SeriesInstanceUID + 1 ) will read the stream up to and including series instance UID, leaving the stream positioned at the next element). The original implementation of this interface, and the default behaviour are preserved for backward compatibility.

@gunterze I'm interested to hear your thoughts on this approach. The problem we are trying to solve is as that we have an input stream positioned at the header of the pixel data element and would like to use DicomInputStream to read only the pixel data, leaving the underlying stream positioned at the beginning of the next tag. Essentially we want to use Tag.PixelData as the stop tag, but would like the read be inclusive of the stop tag.

An alternative interface that would satisfy this objective might look something like:

public Attributes readDataset(int len, int stopTag, boolean stopInclusive);

This approach would also satisfy our immediate need, but would still require that the client somehow predict what elements are present in the stream; which may be too restrictive. For example if we call readDataset( -1, Tag.PixelData ) but the object does not contain pixel data, then the whole stream will be read and there is no opportunity to unread those tags with values larger than 7FE00010, if present:

readDataset( -1, Tag.PixelData );
readDataset( -1, Tag.DigitalSignaturesSequence ); // too late

### DIFF
--- a/dcm4che-core/src/main/java/org/dcm4che3/io/DicomInputStream.java
+++ b/dcm4che-core/src/main/java/org/dcm4che3/io/DicomInputStream.java
@@ -79,8 +79,6 @@ public class DicomInputStream extends FilterInputStream
 
     public enum IncludeBulkData { NO, YES, URI }
 
-    public enum StopTagMatchingStrategy { ExactMatch, MatchOrExceed }
-
     private static final Logger LOG =
         LoggerFactory.getLogger(DicomInputStream.class);
 
@@ -134,7 +132,6 @@ public class DicomInputStream extends FilterInputStream
     private String blkURI;
     private FileOutputStream blkOut;
     private long blkOutPos;
-    private StopTagMatchingStrategy stopStrategy = StopTagMatchingStrategy.ExactMatch;
 
     public DicomInputStream(InputStream in, String tsuid) throws IOException {
         super(in);
@@ -293,14 +290,6 @@ public class DicomInputStream extends FilterInputStream
         return fileMetaInformation;
     }
 
-    public void setStopTagMatchingStrategy(StopTagMatchingStrategy stopStrategy) {
-        this.stopStrategy = stopStrategy;
-    }
-
-    public StopTagMatchingStrategy getStopTagMatchingStrategy() {
-        return stopStrategy;
-    }
-
     public final int level() {
         return itemPointers.length;
     }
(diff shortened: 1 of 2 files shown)


hczedik:
Nice proposal. I also suggested that to @gunterze some time ago already, because we would have similar use cases.

hczedik:
I don't think the different strategies are necessary. I would go for a much simpler approach similar to the second you proposed. Add a new method which simply implements your "MatchOrExceed" strategy. I would have actually expected that behavior from the existing readDataset(int, int) method, but I agree that we have to keep the current behavior of that for backwards compatibility.

mike-mcfarland:
Thanks for your feedback Hermann, agreed simpler is better here. I've pushed an alternative implementation that does not incorporate stop strategies, but that I believe still satisfies the use case we are striving for.

Also, I know this isn't the right forum but I'd also be very interested in timelines for any planned upcoming releases if you have such; I did not see any target dates in the dcm4che forum or the roadmap here: http://www.dcm4che.org/jira/browse/LIB/?selectedTab=com.atlassian.jira.jira-projects-plugin:roadmap-panel

Thanks again!

greg-knox:
If the stop tag is a private tag, should we read to the end of the group as we cannot make any assumptions about how the tags would be resolved for the private creators?

For example: 

0x77770010 -> "PrivateCreator1"
0x77770011 -> "PrivateCreator2"
0x77771010 -> "Tag for Private creator 1"
0x77771110 -> "Tag for Private creator 2"

If the stop tag is given as 0x77771110, it's uncertain whether the consumer would want:

- All 4 tags returned
- Just "PrivateCreator1" and "Tag for Private creator 1" returned
- Just "PrivateCreator2" and "Tag for Private creator 2" returned

mike-mcfarland:
@greg-knox I believe the current approach is okay even with private stop tags. I would suggest that in your example above, the consumer should expect all four tags to be included in the returned dataset (when includeStopTag=true), even though they might only be interested in PrivateCreator2. The real question is, how can the consumer know to specify 0x77771110?

Ideally when specifying a private stop tag, or private tag path for that matter, one should have the ability to provide the element group, private creator ID and element byte instead of specifying an integer outright. For example:

dis.readDataset( -1, new PrivateStopTag( 0x7777, "PrivateCreator2", 0x10 ), false);

Without an interface like this, one must somehow know in advance which element range was reserved for their private group. If an interface like this were available, then my assumption as a consumer would be that the underlying stream is read up to the position of the 7777XX10 element number that correspond to "PrivateCreator2" and that the returned dataset would include all of the other group 7777 private creators' elements encountered up to that point in the stream.

greg-knox:
"Ideally when specifying a private stop tag, or private tag path for that matter, one should have the ability to provide the element group, private creator ID and element byte instead of specifying an integer outright. For example:

dis.readDataset( -1, new PrivateTag( 0x7777, "PrivateCreator2", 0x10 ), false);"

This I agree with. Having separate logic in place for private tags is important as treating them the same as public tags can easily lead to behaviour that the consumer does not expect.

hczedik:
Solved differently on master in the meantime.

PR #38, opened 2016-11-05
StowRS: Minor fixes for formatting / clarity.

davegwatson:
Nothing in this change should affect behavior (aside from minor changes
to error output when incorrect command line args are given).

- Many minor formatting tweaks to aid readability.
- Moved StowMetaDataType from test/ to main/; main code should not
  depend on testing code. Left old enum as deprecated so any code using
  it will not break.
- Command-line parsing is restructured to be more consistent.


Hesham-Elbadawi:
Merged.
Thanks David.

hczedik:
Hesham! Did you actually test this before merging?
The call to instance.stow(); has been removed. The stowrs tool is NOT doing ANYTHING any more since this change.
I'll revert the merge and re-open this pull request.

@davegwatson Can you please fix the problem? We can then review the pull request again.

PR #98, opened 2017-05-08
Update Element Dictionary to DICOM 2017b

gunterze:
dcm4che/dcm4chee-arc-light#684


gunterze:
The (misleading) branch name `dcm4chee-arc-light#669` - instead `dcm4chee-arc-light#684` - was caused by a copy/paste error

PR #115, opened 2017-06-12
Artefacts in decompressed RLE 16bit multi-frame images in all frames except the first

gunterze:
s. dcm4che/dcm4chee-arc-light#778


dcm4che-imageio-rle/src/main/java/org/dcm4che3/imageio/plugins/rle/RLEImageReader.java
```diff
@@ -242,6 +243,7 @@ private void read(byte[][] bands) throws IOException {
 
     private void read(short[] data) throws IOException {
         readRLEHeader(2);
+        Arrays.fill(data, (short) 0);
```
hczedik: Hi Gunter,
thanks for the fix! Shouldn't the null filling also be done in the read(byte[][] bands) method?

gunterze: For 8 bit data, there is no bitwise or with the initial value

PR #118, opened 2017-06-20
dcm4che/dcm4chee-arc-light#799 :  dcm4che-tools-findscu: query params mentioned in dicom binary file or xml file is unused if specified after -m, -i or -r flags

vrindanayak:
dcm4che/dcm4chee-arc-light#799

### DIFF
--- a/dcm4che-tool/dcm4che-tool-findscu/src/main/java/org/dcm4che3/tool/findscu/FindSCU.java
+++ b/dcm4che-tool/dcm4che-tool-findscu/src/main/java/org/dcm4che3/tool/findscu/FindSCU.java
@@ -46,6 +46,7 @@ import java.io.OutputStream;
 import java.security.GeneralSecurityException;
 import java.text.DecimalFormat;
 import java.text.MessageFormat;
+import java.util.ArrayList;
 import java.util.EnumSet;
 import java.util.List;
 import java.util.ResourceBundle;
@@ -154,6 +155,11 @@ public class FindSCU {
     private File xsltFile;
     private Templates xsltTpls;
     private OutputStream out;
+
+    private List<String> argList = new ArrayList<String>();
+    private List<String> returnKeys = new ArrayList<String>();
+    private List<String> matchingKeys = new ArrayList<String>();
+    private List<String> iKeys = new ArrayList<String>();
 
     private Association as;
     private final AtomicInteger totNumMatches = new AtomicInteger();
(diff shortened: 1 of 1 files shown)


dcm4che-tool/dcm4che-tool-findscu/src/main/java/org/dcm4che3/tool/findscu/FindSCU.java
```diff
@@ -437,12 +443,30 @@ private static void configureCancel(FindSCU main, CommandLine cl) {
...
+        addFilesAsArgs(main.argList, cl.getOptionValues("r"), main.returnKeys);
+        addFilesAsArgs(main.argList, cl.getOptionValues("m"), main.matchingKeys);
+        addFilesAsArgs(main.argList, cl.getOptionValues("i"), main.iKeys);
+    }
+
+    private static void addFilesAsArgs(List<String> argList, String[] optVals, List<String> keys) {
+        if (optVals != null)
+            for (String key : optVals) {
+                File f = new File(key);
+                if (f.exists())
```
hczedik: Hi Vrinda,
this logic seems too complex. Why do you have to check if the argument is a file?
If option `-r` `-m` or `-i` is used, then then arguments following those options won't be a file, so e.g.:
```
findscu -i PatientName -i ModalitiesInStudy dicomfile.dcm
```
PatientName and ModalitiesInStudy are obviously not files, because they follow `-i`. On the other hand `dicomfile.dcm` is a file.

hczedik: Ah, I think I understand the problem now.
Maybe we should create these options with
```
.hasArg()
```
instead of
```
.hasArgs()
```
to avoid this ambigouity?

Just imagine the confusion should someone by conincidence have a file called e.g. "PatientName" in his working directory. Your current implementation is unpredictable.

gunterze: .hasArg() would avoid the ambiguity but break the current behavior to do not need to repeat -r -m or -i for specifying multiple arguments for the option. I would vote for keeping the current behavior, and investigate the effort to support the common CLI convention using -- to separate option arguments from program parameters

gunterze: Solved by #126

PR #120, opened 2017-06-26
dcm4che-tool-dcmdir: import records from CSV file into DICOMDIR

vrindanayak:
dcm4che/dcm4chee-arc-light#736

### DIFF
--- a/dcm4che-tool/dcm4che-tool-dcmdir/src/main/java/org/dcm4che3/tool/dcmdir/DcmDir.java
+++ b/dcm4che-tool/dcm4che-tool-dcmdir/src/main/java/org/dcm4che3/tool/dcmdir/DcmDir.java
@@ -47,6 +47,7 @@ import java.text.MessageFormat;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.ResourceBundle;
+import java.util.regex.Pattern;
 
 import org.apache.commons.cli.CommandLine;
 import org.apache.commons.cli.OptionGroup;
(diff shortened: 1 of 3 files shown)


dcm4che-tool/dcm4che-tool-dcmdir/src/main/java/org/dcm4che3/tool/dcmdir/DcmDir.java
```diff
@@ -230,7 +269,48 @@ public static void main(String[] args) {
...
+            for (int i = 0; i < headers.length; i++) {
+                String header = headers[i].replaceAll(quote, "");
+                int tag = DICT.tagForKeyword(header);
+                tags.add(tag);
+                vrs.add(DICT.vrOf(tag));
+            }
+
+            String nextLine;
+            while((nextLine = br.readLine()) != null) {
+                String escq = new StringBuilder().append("\\").append(quote).toString();
```
hczedik: Performance: Build this string only once before the while loop. Same for the next line. (String regex).

(By the way: Why use a StringBuilder here, but not for the next line? I wouldn't use a StringBuilder at all in this particular case. Java automatically replaces such a type of string concatenation by a StringBuilder in the compiled bytecode.)

dcm4che-tool/dcm4che-tool-dcmdir/src/main/java/org/dcm4che3/tool/dcmdir/DcmDir.java
```diff
@@ -230,7 +269,48 @@ public static void main(String[] args) {
...
+                String regex = delim + "(?=(?:[^" + escq + "]*" + escq + "[^" + escq + "]*" + escq + ")*[^" + escq + "]*$)";
+                String[] values = nextLine.split(regex, -1);
```
hczedik: Performance: Better use Pattern.compile(..).split(...) and do the Pattern.compile(...) outside of the loop.

PR #125, opened 2017-06-30
Support Range Matching by Attributes.matches(Attributes keys,..)

gunterze:
dcm4che/dcm4chee-arc-light#830

### DIFF
--- a/dcm4che-core/src/main/java/org/dcm4che3/data/Attributes.java
+++ b/dcm4che-core/src/main/java/org/dcm4che3/data/Attributes.java
@@ -790,7 +790,7 @@ public class Attributes implements Serializable {
 
     private static String[] toStrings(Object val) {
         return (val instanceof String) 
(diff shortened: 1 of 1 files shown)


dcm4che-core/src/main/java/org/dcm4che3/data/Attributes.java
```diff
@@ -2943,9 +2943,19 @@ private boolean matches(String privateCreator, int tag, VR vr,
...
                 keyVal = new PersonName(keyVals[0]).toString();
-    
+            switch (vr) {
+                case PN:
+                    keyVal = new PersonName(keyVals[0]).toString();
+                    break;
+                case DA:
+                case DT:
+                case TM:
+                    dateRange = toDateRange(keyVal, vr);
```
gunterze: added the break and removed some redundant lines...

PR #149, opened 2017-08-03
Thread deadlock on receive of A-RELEASE-RQ after P-DATA-TF with pending PDV

gunterze:
dcm4che/dcm4chee-arc-light#894

Applies #147 to generic-config (=previous master)
Replaces #114


alex-hoermandinger:
Cleaner Fix for the same problem previously fixed by commit 49f22b688a47d65d39029612dd0b42a4b99553c1

Issue #151, opened 2017-08-04
InputStream remains open after "Not a Dicom Stream" Exception


alej4ndro:
Hi everybody,

There is a problem using dcm4che tools, where these are applied to a folder of dicom objects and in that folder also exists some non-dicom files. For example, the command:

dcm2jpg <dcmfolder> <jpgfolder>

DicomFiles class scans and identifies dicom files inside the "dcmfolder". However, when it founds a non-dicom file, the line: 

[DicomInputStream in = new DicomInputStream(f);](https://github.com/dcm4che/dcm4che/blob/master/dcm4che-tool/dcm4che-tool-common/src/main/java/org/dcm4che3/tool/common/DicomFiles.java#L104)

Throws a "Not DICOM Stream" exception comming from "guessTransferSyntax()". This exception throws after a "new FileInputStream(f)" is created, and before finalyzing the instantiation of the DicomInputStream. As a consequence, the var "in" from "in = new DicomInputStream(f);" remains null when the code execution reach the "finally" block due to the described exception. Then, in the "finally" block Safe.close(in) is called. However, "in" is null and Safe.close does nothing.  Thus, [the FileInputStream created inside "new DicomInputStream(File f)"](https://github.com/dcm4che/dcm4che/blob/master/dcm4che-core/src/main/java/org/dcm4che3/io/DicomInputStream.java#L151)  remains open indefinitely, and later on I have no option to get back that stream to properly close it. Please, solve this issue by trying-catch the guessTransfersyntax function in the DicomInputStream builder. if some problem occurs guessing the transfer syntax, please close the FileInputStream in the catch block (inside the DicomInputStream(InputStream in) builder) and then throw the same exception. This way the bug can be solved and the application could keep working without changes.

Glad to contribute. Hope this help.

Best regards,
Alej4ndro.

hczedik:
Hi Alej4ndro,
thanks for reporting this.
Pull request are welcome!

PR #153, opened 2017-08-08
DicomInputStream builder bug fixed

alej4ndro:
Now it tries to close the FileInputStream if something was wrong guessing the transfer syntax.
Related to #151


gunterze:
Applied with minimal modifications:  ef86b65

Issue #152, opened 2017-08-04
Messages in the spool directory are not send after restart if spoolDirectory property in AuditLogger is not set


markusgr:
The method org.dcm4che3.net.audit.AuditLogger#getNumberOfQueuedMessages returns 0 if property spoolDirectory is not set. This causes an issue if the system is restarted and messages are in the queue. Only new messages will be sent.

PR #154, opened 2017-08-08
Bugfix for Issue #152

markusgr:
Messages in the spool directory are not send after restart if spoolDirectory property in AuditLogger is not set #152


gunterze:
Fixed by  9857fb3

PR #162, opened 2017-08-28
Generic config

carreraa8:


greg-knox:
Pulling in more than intended - a separate pull request will be opened for specifically this change only: https://github.com/dcm4che/dcm4che/commit/2388d9729e7d40d66791ba0d1b614759a1c30276

PR #163, opened 2017-09-05
Added scope test to junit dependencies…

patrick-peer:
…except for some tool artifacts, that have test classes in src/main/java. I also removed the versions in dcm4che-audit/pom.xml and dcm4che-soundex/pom.xml. I made sure 'mvn clean install', 'mvn clean compile' and 'mvn clean test' still work.

I did this to make compiled binaries not depend on and/or include junit. Before, this forced users of the library to include/depend on junit for their binaries as well.

As a sidenote, I would recommend to move the test classes in dcm4che-tool/dcm4che-tool-findscu, dcm4che-tool/dcm4che-tool-storescu and dcm4che-tool/dcm4che-tool-storescu from src/main/java to src/test/java to remove the compile time dependency of junit for those artifacts. Do you agree?


hczedik:
:thumbsup:
I agree that the Test classes within the tools shouldn't be there. I would delete them completely. The generic-config branch contains newer versions of these test classes, and we plan to extract them into an own project.

Issue #159, opened 2017-08-24
Unable to parse DICOM instance with sequence of VR=UN containing explicit VR tags


justinfalk:
dcm4che3 (v3.3.8) will not parse a DICOM instance with a non-compliant sequence item containing explicit VR tags inside a sequence tag having VR of UN.  
Other toolkits such as dcm4che14, dcm4che2 and dcmtk are able to parse the instance.   

In the attached example, the length of the first sequence item for tag (0010,2202) is 42, but when parsing the first tag dcm4che3 interprets the explicit VR of "UL" plus the next two bytes as a length of 281685 (it's not expecting the explicit VR entry).  dcm4che3 attempts to read the next 281685 bytes and then get confused as it attempts to find the next tag from somewhere within the pixel data.  The end result is an "Unrecogniced VR code” warning followed by an EOF exception.  The other toolkits mentioned above fail to parse the sequence item’s incorrectly encoded tags but still are able to parse the subsequent tags and render the instance. 

![screen shot 2017-08-24 at 10 04 18 am](https://user-images.githubusercontent.com/16879499/29684644-5de8e318-88d8-11e7-8a6f-17dc605a24fb.png)

Obviously the root cause is that the instance is non-conformant, but it seems that this situation could be handled better by the toolkit.  I see that there is a “decodeUNWithIVRLE” option in DicomInputStream.  Setting this to false allows parsing of this instance, however, one often won't have a priori knowledge of the instance containing this defect.  It seems that the toolkit could detect child tags with length > the containing item length and stop trying to parse the sequence item.  

Stack trace:  

```
2017-08-24 14:20:02.723  WARN 11200 --- [io-18080-exec-1] org.dcm4che3.data.VR                     : Unrecogniced VR code: {0}H - treat as UN
2017-08-24 14:20:02.727  WARN 11200 --- [io-18080-exec-1] org.dcm4che3.io.DicomInputStream         : IOException during read of (58D0,CF50) #-1095488214 @ 282709

java.io.EOFException: null
    at org.dcm4che3.io.DicomInputStream.readValue(DicomInputStream.java:725)
    at org.dcm4che3.io.DicomInputStream.readValue(DicomInputStream.java:555)
    at org.dcm4che3.io.DicomInputStream.readAttributes(DicomInputStream.java:516)
    at org.dcm4che3.io.DicomInputStream.readDataset(DicomInputStream.java:444)
```

[...]

PR #164, opened 2017-09-07
Prevent read of tag beyond specified len (issue #159) 

justinfalk:
While reading attributes of a defined length, skip tags that exceed to boundaries of the provided length.  This can happen with incorrectly encoded instance (for example, explicitly encoded tags with UN sequence item).


gunterze:
Solved #159 by  642eef4

Issue #183, opened 2017-11-08
DicomImageReader reads/skips entire DICOM image input stream, instead of just reading to pixel data


wayfarer3130:
The DicomImageReader implementation reads the full fragment object on the pixel data, AND reads full large non-pixel data objects instead of just parsing up to the pixel data/basic offset table, and then using the basic offset table to find positions/read the rest of the pixel data only as required.

wayfarer3130:
The generic-config branch of this has an updated version that supports this, as well as supporting the master branch stream reading using the dicom input streams.  It passes both the Master and generic-config unit tests.

PR #185, opened 2017-11-15
#183 DicomImageReader reads entire DICOM image input stream instead o…

wayfarer3130:
I've changed the dicom image reader to only read the dicom image input stream up to the pixel data tag + basic offset table so that it doesn't need to have read/skipped anything after the desired image before being able to decode it.  That makes it far faster on large multi-frames, and makes the behaviour similar to the dicom input stream behaviour.


dcm4che-core/src/main/java/org/dcm4che3/util/ByteUtils.java
```diff
@@ -359,4 +359,11 @@ private static void swapLastFirst(byte[] b1, byte[] b2) {
         b1[last] = t;
     }
 
+    public static byte[] intsToBytesLE(int... values) {
```
wayfarer3130: This is needed for the basic offset table, which is always LEI

dcm4che-core/src/main/java/org/dcm4che3/data/BulkData.java
```diff
@@ -227,4 +228,63 @@ public static Value deserializeFrom(ObjectInputStream ois)
             StringUtils.maskEmpty(ois.readUTF(), null),
             ois.readBoolean());
     }
+
+    @Override
+    public boolean equals(Object obj) {
```
wayfarer3130: I wanted some comparison capability to make sure even after update the values compare correctly to originally created values.

dcm4che-imageio/src/main/java/org/dcm4che3/imageio/plugins/dcm/DicomImageReader.java
```diff
@@ -161,23 +165,26 @@ public void setInput(Object input, boolean seekForwardOnly,
...
+    private void initPixelDataFromAttributes(Attributes ds) {
         VR.Holder holder = new VR.Holder();
         Object value = ds.getValue(Tag.PixelData, holder);
         if (value != null) {
             pixelDataVR = holder.vr;
             if (value instanceof BulkData) {
                 pixelData = (BulkData) value;
                 pixelDataLength = pixelData.length();
-            } else { // value instanceof Fragments)
+            } else if( value instanceof byte[] ) {
```
wayfarer3130: This is needed to handle decoding uncompressed full Attributes instance data, added an integration test for this type of data.

dcm4che-imageio/src/main/java/org/dcm4che3/imageio/plugins/dcm/DicomImageReader.java
```diff
@@ -187,8 +194,12 @@ private void initPixelData(Attributes ds) {
     private void initPixelDataFile() {
         if (pixelData != null)
             pixelDataFile = pixelData.getFile();
-        else if (pixelDataFragments != null && pixelDataFragments.size() > 1)
-            pixelDataFile = ((BulkData) pixelDataFragments.get(1)).getFile();
+        else if (pixelDataFragments != null && pixelDataFragments.size() > 1) {
+            Object frag = pixelDataFragments.get(1);
+            if( frag instanceof BulkData ) {
```
wayfarer3130: Might be byte[] data

dcm4che-imageio/src/main/java/org/dcm4che3/imageio/plugins/dcm/DicomImageReader.java
```diff
@@ -277,7 +293,7 @@ public ImageReadParam getDefaultReadParam() {
     }
 
     @Override
-    public IIOMetadata getStreamMetadata() throws IOException {
+    public DicomMetaData getStreamMetadata() throws IOException {
```
wayfarer3130: Convenient to specialize this so it already has the right type.

dcm4che-imageio/src/main/java/org/dcm4che3/imageio/plugins/dcm/DicomImageReader.java
```diff
@@ -429,8 +454,14 @@ public BufferedImage read(int frameIndex, ImageReadParam param)
...
-    @SuppressWarnings("resource")
-    private ImageInputStream iisOfFrame(int frameIndex) throws IOException {
+    /** Generate an image input stream for the given frame, -1 for all frames (video, multi-component single frame)
+     * Does not necessarily support the length operation without seeking/reading to the end of the input.
+     * 
+     * @param frameIndex
+     * @return
+     * @throws IOException
+     */
+    public ImageInputStream iisOfFrame(int frameIndex) throws IOException {
```
wayfarer3130: Making this public allows reading the raw bytes from the input, which can be useful if the actual image encoding isn't being changed, but a file is just being written from the original data.

dcm4che-imageio/src/main/java/org/dcm4che3/imageio/plugins/dcm/DicomImageReader.java
```diff
@@ -587,22 +618,83 @@ private void readMetadata() throws IOException {
                 pixelDataLength = dis.length();
                 if (pixelDataLength == -1)
                     epdiis = new EncapsulatedPixelDataImageInputStream(dis, ds.getInt(Tag.NumberOfFrames, 1));
+            } else {
```
wayfarer3130: I will have another review/issue to handle the rest of the post pixel data changes, but this example handles it if the object happens to not be an image containing object.

dcm4che-imageio/src/main/java/org/dcm4che3/imageio/plugins/dcm/DicomImageReader.java
```diff
@@ -587,22 +618,83 @@ private void readMetadata() throws IOException {
...
+    /** Creates an offset/length table based on the frame positions */
+    public static void generateOffsetLengths(Fragments pixelData, int frames, byte[] basicOffsetTable, long start) {
+        long lastOffset = 0;
+        BulkData lastFrag = null;
+        for(int frame=0; frame<frames; frame++) {
+            long offset = frame>0 ? 1 : 0;
+            int offsetStart = frame*4;
+            if( basicOffsetTable.length>=offsetStart+4 ) {
+                offset = ByteUtils.bytesToIntLE(basicOffsetTable, offsetStart);
+                if( offset!=1 ) {
```
wayfarer3130: Allow 1 values to indicate unknown yet basic offset table values - that allows writing an image, then going back and fixing up the basic offset table, will be used in a later code review.

dcm4che-imageio/src/test/java/org/dcm4che3/imageio/dcm/TestDicomImageReader.java
```diff
@@ -61,11 +76,17 @@
  */
 public class TestDicomImageReader {
 
-    ImageReader reader;
+    private static final String CPLX_P02 = "cplx_p02.dcm";
+    private static final String NM_MF = "NM-MONO2-16-13x-heart";
+    private static final String NM_MF_CHECKSUM = "B2813DA2FE5B79A1B3CAF18DBD25023E2F84D4FE";
```
wayfarer3130: Added some checksums to make sure we get the right pixel data.

hczedik:
Looks good to me! :thumbsup:

dcm4che-imageio/src/main/java/org/dcm4che3/imageio/plugins/dcm/DicomImageReader.java
```diff
@@ -95,6 +96,8 @@
 public class DicomImageReader extends ImageReader {
 
     private static final Logger LOG = LoggerFactory.getLogger(DicomImageReader.class);
+
+	public static final String POST_PIXEL_DATA = "postPixelData";
```
hczedik: Please use only spaces for intendation. No tabs.

dcm4che-imageio/src/main/java/org/dcm4che3/imageio/plugins/dcm/DicomImageReader.java
```diff
@@ -775,6 +810,46 @@ private void checkIndex(int frameIndex) {
             throw new IllegalStateException(
                     "input stream position already after requested frame #" + (frameIndex + 1));
     }
+    
+    /** Reads post-pixel data tags, will skip past any remaining images (which may be very slow), and
+     * add any post-pixel data information to the attributes object.
+     * NOTE: This read will read past image data, and may end up scanning/seeking through multiframe or video data in order tofind the
```
hczedik: Typo: Space missing in this sentence.

dcm4che-imageio/src/main/java/org/dcm4che3/imageio/plugins/dcm/DicomImageReader.java
```diff
@@ -775,6 +810,46 @@ private void checkIndex(int frameIndex) {
...
+     * post pixel data.  This may be slow.
+     *
+     * Replaces the attributes object with a new one, thus is thread safe for other uses of the object.
+     *
+     * Does NOT work with LEI compressed data.
```
hczedik: Please write out "LEI" -> Implicit VR Little-endian

dcm4che-imageio/src/main/java/org/dcm4che3/imageio/plugins/dcm/DicomImageReader.java
```diff
@@ -775,6 +810,46 @@ private void checkIndex(int frameIndex) {
...
+     */
+    public Attributes readPostPixeldata() throws IOException {
+    	if( frames==0 ) return metadata.getAttributes();
+    	
+        if( dis!=null ) {
+        	if( flushedFrames > frames ) {
```
hczedik: Please only use spaces for intendation. No tabs.

dcm4che-imageio/src/test/java/org/dcm4che3/imageio/dcm/TestDicomImageReader.java
```diff
@@ -95,6 +100,59 @@ public void tearDown() throws Exception {
...
+            assertThat(r).isNotNull();
+        }
+        Attributes postAttr = reader.readPostPixeldata();
+        assertThat(postAttr.getString(postPixelCreator,postPixelTag)).isEqualTo("Value");
+        if( input instanceof ImageInputStream ) {
+            // Must not throw an exception
+            reader.readRaster(0, null);
+        } else {
+        	try {
+        		reader.readRaster(0, null);
```
hczedik: Same here: please only spaces for indentation.

dcm4che-imageio/src/main/java/org/dcm4che3/imageio/plugins/dcm/DicomImageReader.java
hczedik: How will it behave when I try to use it with LEI compressed data?

wayfarer3130:
It will throw exceptions, although that isn't any different than the
previous version which would throw exceptions trying to read ​even images
using an image input stream on a compressed LEI stream.

I will update the comment to indicate it works for LEI compressed if you
use an input stream instead.  Might modify the code to explicitly handle
that cleanly.  It isn't just that part that fails - the bulk data objects
also fail to work on compressed.

I also think there is a fundamental issue with non-multiple of 8 bit
images, not introduced in my code but not fixed yet either.

Do you have an eclipse java style sheet I can import, or at least a
description of the formatting?  There used to be one on the website, but I
can't find it any longer.

Thanks,
Bill.

hczedik:
Hi Bill,

I think the "Java Conventions [built-in]" profile of Eclipse should be fine.

wayfarer3130:
With some tweaks I got it to remove spaces and have updated that, as well
as the javadoc comments you requested.

gunterze:
Sorry, it does not seem that I can take the time too review the changes in detail in near future. I just did some testing, which does not show up any defect.

PR #190, opened 2017-11-27
Ensure we pick a compressed transfer syntax when available

robedwardsagfa:
(cherry picked from commit ac9d587658412f06d1214a1cabda7408b76c79ca)

### DIFF
--- a/dcm4che-net/src/main/java/org/dcm4che3/net/ApplicationEntity.java
+++ b/dcm4che-net/src/main/java/org/dcm4che3/net/ApplicationEntity.java
@@ -514,14 +514,16 @@ public class ApplicationEntity implements Serializable {
                    PresentationContext.ABSTRACT_SYNTAX_NOT_SUPPORTED,
                    rqpc.getTransferSyntax());
 
-       for (String ts : rqpc.getTransferSyntaxes())
-           if (tc.containsTransferSyntax(ts)) {
-               byte[] info = negotiate(rq.getExtNegotiationFor(as), tc);
-               if (info != null)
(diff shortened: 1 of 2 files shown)


robedwardsagfa:
One other thing I should point out is this change replicates what was done previously in dcm4che2.  Allowing a prioritized ordering seemed to have been inadvertently removed with the migration to dcm4che3.

gunterze:
Rejected, because an application which reads the configuration from a LDAP backend according [PS3.15 H.1. Application Configuration Management Profile](http://dicom.nema.org/medical/dicom/current/output/html/part15.html#sect_H.1) may get the list of Transfer Syntaxes of a particular Transfer Capability in different order for each read, and therefore would behave nondeterministic.

hczedik:
Hi Gunter,
if the order of the list of Transfer Syntaxes returned by LDAP is nondeterministic, couldn't we simply sort them (e.g. alphabetically) then so that a predictable sorting is ensured?

gunterze:
But selecting always the literal first Transfer Syntax is less sensitive than the current behavior to select the first in the offered list in the Presentation Context.

hczedik:
Yes, but then it is better configurable for those using a configuration backend that supports setting a fixed order of transfer capabilities.
Is LDAP really always unsorted? Seems like a disadvantage.

gunterze:
Behavior should not depend on the configuration backend!

You may e.g. define an additional attribute on Device, AE and/or Transfer Capability level "Preferred Transfer Syntax" with a list of Transfer Syntax UIDs stored in LDAP  in a single value using any delimiter character...

gunterze:
Fixed by #218 15ccd97

PR #192, opened 2017-11-28
IEI-49757 Compression of frames on storage results in Compressor$CompressedFrame on finalizer queue

mfarooq2000:
Compression of frames on storage results in Compressor$CompressedFrame
on finalizer queue


mfarooq2000:
@hczedik please review. Thanks

Issue #200, opened 2017-12-15
Parsing large sequences is O(n^2) operation because of itemPointers lookup


wayfarer3130:
In the generic-config and rc81 branches, parsing large sequences is an O(n^2) operation because the item pointers generated are done once per nested sequence item, and at each sequence level do an O(n) operations - thus, the total cost is O(nmk) for a sequence hierarchy of size n, m and k respectively - in general this is 2 levels only, so saying it is O(n^2) as a general case.

PR #201, opened 2017-12-15
Changes for issue #200

wayfarer3130:
Hi Greg - this is the change for the slow sequence parsing issue I was seeing.    I had seen a couple of other very small issues when I was working on this that have been included and which I will comment on.

### DIFF
--- a/dcm4che-core/src/main/java/org/dcm4che3/data/Attributes.java
+++ b/dcm4che-core/src/main/java/org/dcm4che3/data/Attributes.java
@@ -38,22 +38,32 @@
 
 package org.dcm4che3.data;
 
+import java.io.IOException;
+import java.io.ObjectInputStream;
+import java.io.ObjectOutputStream;
+import java.io.Serializable;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Date;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.TimeZone;
+import java.util.regex.Pattern;
+
 import org.dcm4che3.data.IOD.DataElement;
 import org.dcm4che3.data.IOD.DataElementType;
 import org.dcm4che3.io.BulkDataDescriptor;
 import org.dcm4che3.io.DicomEncodingOptions;
 import org.dcm4che3.io.DicomInputStream;
 import org.dcm4che3.io.DicomOutputStream;
-import org.dcm4che3.util.*;
+import org.dcm4che3.util.ByteUtils;
+import org.dcm4che3.util.DateUtils;
+import org.dcm4che3.util.Deidentifier;
+import org.dcm4che3.util.StringUtils;
+import org.dcm4che3.util.TagUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-
-import java.io.IOException;
-import java.io.ObjectInputStream;
-import java.io.ObjectOutputStream;
-import java.io.Serializable;
-import java.util.*;
-import java.util.regex.Pattern;
 
 /**
  * @author Gunter Zeilinger <gunterze@gmail.com>
@@ -1951,8 +1961,10 @@ public class Attributes implements Serializable {
     }
 
     private Object set(String privateCreator, int tag, VR vr, Object value) {
-        if (vr == null)
-            throw new NullPointerException("vr");
+        if (vr == null) {
+            LOG.warn("Provided NULL vr for {}:{} value {}", privateCreator,Integer.toHexString(tag), value);
+            vr = VR.UN;
+        }
 
         if (privateCreator != null) {
             int creatorTag = creatorTagOf(privateCreator, tag, true);
@@ -2199,6 +2211,12 @@ public class Attributes implements Serializable {
     }
 
     public boolean addWithoutBulkData(Attributes other, BulkDataDescriptor descriptor) {
+        List<ItemPointer> itemPointers = new ArrayList<>();
+        itemPointers.addAll(Arrays.asList(other.itemPointers()));
+        return addWithoutBulkData(other,descriptor,itemPointers);
+    }
+    
+    private boolean addWithoutBulkData(Attributes other, BulkDataDescriptor descriptor, List<ItemPointer> itemPointers) {
         final boolean toggleEndian = bigEndian != other.bigEndian;
         final int[] tags = other.tags;
         final VR[] srcVRs = other.vrs;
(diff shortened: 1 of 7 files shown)


dcm4che-core/src/main/java/org/dcm4che3/data/Attributes.java
```diff
@@ -1951,8 +1961,10 @@ public Fragments newFragments(String privateCreator, int tag, VR vr,
     }
 
     private Object set(String privateCreator, int tag, VR vr, Object value) {
-        if (vr == null)
-            throw new NullPointerException("vr");
+        if (vr == null) {
```
wayfarer3130: The file I was testing with had a 0 encoded instead of an UN code, which caused vr=null here.

dcm4che-core/src/main/java/org/dcm4che3/data/Attributes.java
```diff
@@ -2199,6 +2211,12 @@ public boolean addSelected(Attributes other, String privateCreator, int tag) {
     }
 
     public boolean addWithoutBulkData(Attributes other, BulkDataDescriptor descriptor) {
+        List<ItemPointer> itemPointers = new ArrayList<>();
+        itemPointers.addAll(Arrays.asList(other.itemPointers()));
```
wayfarer3130: This particular item is an O(n) call, but the addWithoutBulkData should be called externally very occasionally, so that shouldn't be too expensive overall.

dcm4che-core/src/main/java/org/dcm4che3/data/Attributes.java
```diff
@@ -2236,30 +2253,31 @@ public boolean addWithoutBulkData(Attributes other, BulkDataDescriptor descripto
...
 
-    private void setWithoutBulkData(String privateCreator, int tag, Sequence seq,
+    private void setWithoutBulkData(List<ItemPointer> itemPointers, String privateCreator, int tag, Sequence seq,
                                     BulkDataDescriptor descriptor) {
         Sequence newSequence = newSequence(privateCreator, tag, seq.size());
         for (Attributes item : seq) {
             Attributes newItem = new Attributes(bigEndian, item.size());
             newSequence.add(newItem);
-            newItem.addWithoutBulkData(item, descriptor);
+            itemPointers.add(new ItemPointer(privateCreator, tag, newSequence.size()));
```
wayfarer3130: This add/remove allows the add without bulk data to run the the O(1) version of the lookup.

dcm4che-core/src/main/java/org/dcm4che3/data/Fragments.java
```diff
@@ -87,6 +87,10 @@
     private final VR vr;
     private final boolean bigEndian;
 
+    public Fragments(VR vr, boolean bigEndian, int initialCapacity) {
```
wayfarer3130: This is just a change to make it look more like the master branch because I was testing this it this way, not otherwise used yet, but I plan to make more commits over time to improve agreement between master and the GC branch

dcm4che-core/src/main/java/org/dcm4che3/data/VR.java
```diff
@@ -221,6 +221,10 @@
     }
 
     private static int indexOf(int code) {
+        if( code==0 ) {
```
wayfarer3130: This is the other side of the null/0 valued VR change.

dcm4che-core/src/main/java/org/dcm4che3/io/BulkDataDescriptor.java
```diff
@@ -78,31 +79,19 @@ public boolean isBulkData(String privateCreator, int tag, VR vr, int length,
...
-            case OD:
-            case OF:
-            case OL:
-            case OW:
-            case UN:
-                return length > 64*1024*2;
-                default:
-            }
-            return false;
+            // Don't need any private tags larger than 64k
```
wayfarer3130: I was merging stuff from master and this got included - it doesn't make a huge difference but makes things a bit more consistent.

dcm4che-core/src/main/java/org/dcm4che3/io/DicomInputStream.java
```diff
@@ -703,23 +700,25 @@ private void readSequence(int len, Attributes attrs, int sqtag)
...
-            if (tag == Tag.Item) {
-                handler.readValue(this, seq);
-            } else if (tag == Tag.SequenceDelimitationItem) {
-                if (length != 0)
-                    skipAttribute(UNEXPECTED_NON_ZERO_ITEM_LENGTH);
-                break;
-            } else
-                skipAttribute(UNEXPECTED_ATTRIBUTE);
+        for (int i = 0; (undefLen || pos < endPos) && readItemHeader(); ++i) {
+            addItemPointer(sqtag, privateCreator, i);
```
wayfarer3130: And this set of changes here preserve the pointer data is what makes the change work

dcm4che-core/src/main/java/org/dcm4che3/data/Attributes.java
hczedik: I think this shouldn't be part of this commit. If you want to handle that more lenient, I would suggest to create a separate fix and also open that as a pull request for the master branch.

dcm4che-core/src/main/java/org/dcm4che3/io/BulkDataDescriptor.java
hczedik: This is not part of the master branch! I think it is only part of your pull request, and Gunter will probably not accept this. I think the PIXELDATA BulkDataDescriptor should only remove PixelData. But of course we can always add an additional descriptor, e.g. something like PIXELDATA_AND_LARGE_TAGS.

wayfarer3130:
The pixel data descriptor is for reading the dicom pixel data.  Thus, it
should avoid reading anything too large/not needed for decoding the pixel
data that is large enough to matter.  I'm still working a bit on figuring
out exactly what that should be, for example, this change avoids excluding
large overlays that the DIR needs - it would have been able to exclude a
few more things, but the basic idea is there.  The ONLY use in dcm4che3 is
in the dicom image reader, so I don't feel at this point that it is
justified to create another instance, but rather change this one to meet
the needs of the DIR.

hczedik:
I wouldn't say the BulkDataDescriptor.PIXELDATA is necessarily for reading the pixeldata.
Other applications might use that for reading everything BUT the pixeldata (to not load too much into memory). So I think we should not change that behavior.
Please introduce a new BulkDataDescriptor for your intended behavior.

PR #216, opened 2018-01-16
SyslogProtocolHandler to use his own ThreadPoolExecutor

umberto-cappellini:
I prepared SyslogProtocolHandler to use his own ThreadPoolExecutor instead of the common Device Executor. At the moment the own Executor is commented out in the constructor, so the behavior is exactly as before (using the Device's Executor). In the branch "generic-config" is already uncommented. Maybe should be made configurable.

### DIFF
--- a/dcm4che-net-audit/src/main/java/org/dcm4che3/net/audit/SyslogProtocolHandler.java
+++ b/dcm4che-net-audit/src/main/java/org/dcm4che3/net/audit/SyslogProtocolHandler.java
@@ -45,8 +45,13 @@ import java.net.DatagramPacket;
 import java.net.InetAddress;
 import java.net.Socket;
 import java.net.SocketTimeoutException;
+import java.util.concurrent.Executor;
+import java.util.concurrent.SynchronousQueue;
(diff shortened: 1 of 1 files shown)


dcm4che-net-audit/src/main/java/org/dcm4che3/net/audit/SyslogProtocolHandler.java
```diff
@@ -66,14 +71,34 @@
...
+    private void execute ( Runnable runnable, Executor defaultExecutor ) {
+
+        if (executor!=null) {
+            executor.execute( runnable );
+        }
+        else if (defaultExecutor!=null) {
+            defaultExecutor.execute( runnable );
+        }
+        else
+            throw new IllegalStateException("executor not initalized");
```
hczedik: Typo: initalized -> initialized

gunterze:
Why not just using an own Device for the Audit Record Repository - or what ever uses the syslog server connection - so the Executer configured for that Device is only used from the SyslogProtocolHandler.

PR #281, opened 2018-05-08
Changes to fix issue 245, failing to read the entire image in images …

wayfarer3130:
There were 2 issues, plus a dependency from opencv on the internal code that probably shouldn't have been included.  The first issue was that for multiple part images, we need to pass in -1 for the frame number to indicate read-all.  The second was an EOF over-run issue causing a null pointer.
The dependency was that opencv used the actual number of segments from the internal segmented input to read the response, and that wasn't computed up front, so it was allocating an array size too large.  The work around is to read the size up front, although that is less than optimal as it does a double read.


gunterze:
Successfully tested on ftp://medical.nema.org/MEDICAL/Dicom/DataSets/WG04

gunterze:
Successfully tested on ftp://medical.nema.org/MEDICAL/Dicom/DataSets/WG04

PR #285, opened 2018-05-14
add honor cipher suites

bashims:


bashims:
This feature request requires openjdk-1.8, but it seems that travis job requires openjdk-1.7. What's the best way to go about addressing multi JDK  version support?

Thanks,

-ben

gunterze:
You may use reflection to only invoke `SSLContext.setUseCipherSuitesOrder(boolean)` if the method exists.
You should also take care that LDAP does not guarantee, to preserve the order of multiple values. So you cannot rely on the order of `String[] tlsCipherSuites`!

PR #294, opened 2018-06-05
Be tolerant of nulls for IS and DS values in JSON encoding.

spalte:
When writing out poorly formatted IS and DS values dcm4che replaces
the poorly formated value with a null. The resulting JSON can not
be read back in by dcm4che.

JSONWriter.java
```
            case DS:
                try {
                    gen.write(StringUtils.parseDS(s));
                } catch (NumberFormatException e) {
                    LOG.info("illegal DS value: {} - encoded as null", s);
                    gen.writeNull();
                }
                break;
            case IS:
                try {
                    gen.write(StringUtils.parseIS(s));
                } catch (NumberFormatException e) {
                    LOG.info("illegal IS value: {} - encoded as null", s);
                    gen.writeNull();
                }
                break;
```

### DIFF
--- a/dcm4che-json/src/main/java/org/dcm4che3/json/JSONReader.java
+++ b/dcm4che-json/src/main/java/org/dcm4che3/json/JSONReader.java
@@ -386,7 +386,8 @@ public class JSONReader {
         double[] toDoubles() {
             double[] ds = new double[values.size()];
             for (int i = 0; i < ds.length; i++) {
-                ds[i] = ((Number) values.get(i)).doubleValue();
+                Number value = (Number) values.get(i);
+                ds[i] = value != null ? value.doubleValue() : 0;
             }
             return ds;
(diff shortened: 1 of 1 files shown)


jssuttles:
Also... I'd like to point out that `LOG.info("illegal DS/IS value: {} - encoded as null", s);` spits out in the middle of stdout, which happens to be where the JSON is being generated when using dcm2json. This creates an unparseable JSON object.
My current workaround is to regex this error out. But, I'd rather it be logged somewhere or dealt with in another way.

wtellis:
I'd like to add my vote for this fix.  This issue is causing a lot of problems for one of our apps and the proposed fix seems to address them.

gunterze:
Alternatively fixed by f898e3d #336

spalte:
Thanks!

PR #321, opened 2018-07-23
[IEI-67902] Type 1 attribute (0010,0022) is missing in the other Patient ID sequence (0010,1002)

mfarooq2000:
@robedwardsagfa please review and approve

Cloned from 
Commits on Jan 5, 2018
Type 1 attribute (0010,0022) is missing in the other Patient ID sequ...
by carreraa8 committed on Jan 5


hczedik:
@robedwardsagfa @mfarooq2000 Isn't this something which should also be opened as a pull request towards the master branch?

PR #326, opened 2018-07-25
Type 1 attribute (0010,0022) is missing in the other Patient ID sequence (0010,1002)

mfarooq2000:
@hczedik Please review. Thanks


hczedik:
@mfarooq2000 I think the TypeOfPatientID should only be part of the  OtherPatientIDsSequence. I'm not sure what you use exportPatientIDWithIssuer(.) for (or for what exactly it is intended to be used for, as it does not have any Javadoc), but if we also use it for exporting into the toplevel dataset, then that does not look correct.
@gunterze Can you also have a look at this please?

gunterze:
Alternatively solved by  #338  c86dfb6

mfarooq2000:
@gunterze
Method exportPatientIDWithIssuer(.) is used to create PatientID element on root level and OtherPatientIDsSequence. Agreed TypeOfPatientID should only be part of OtherPatientIDsSequence and is Type 1 attribute, so my suggestion is to have another method which will add TypeOfpatientID with value "TEXT" by default in OtherPatientIDsSequence 

	public Attributes exportOtherPatientIDWithIssuer(Attributes attrs) {
		attrs = exportPatientIDWithIssuer(attrs);
		if (typeOfPatientID == null) {
			attrs.setString(Tag.TypeOfPatientID,VR.CS,"TEXT");
		}
		return attrs;	
    }


Also the check if (typeOfPatientID != null) in hashCode() is causing the method pidsOf(.) to return same pid twice in case same pid is already in OtherPatientIDsSequence since the fact typeOfPatientID doesn't exist in root level and in OtherPatientIDsSequence it has a non-null value.
Updated hashCode method:

    @Override
    public int hashCode() {
        int result = id.hashCode();
        if (identifierTypeCode != null)
            result += identifierTypeCode.hashCode() * 31;
        if (issuer != null)
            result += issuer.hashCode() * 31;
        return result;
    }

mfarooq2000:
@gunterze @hczedik please review when you get chance. Thanks

gunterze:
Your exportOtherPatientIDWithIssuer method breaks the constraint:
```
IDWithIssuer pid = ...
IDWithIssuer.pidOf(pid.exportOtherPatientIDWithIssuer(null)).equals(pid);
```
See no benefit having to use a different method compared to
```
pid.setTypeOfPatientID(TypeOfPatientID.TEXT);
pid.exportPatientIDWithIssuer(null);
```
 
Your proposed change of the hashCode() would not avoid duplicate pids returned by  pidsOf(.), because the equals method still return false, if two pids differs only by their identifierTypeCode.

mfarooq2000:
I agree we can go with 

`pid.setTypeOfPatientID(TypeOfPatientID.TEXT);
pid.exportPatientIDWithIssuer(null);`

The actual problem 
**Type 1 attribute (0010,0022) is missing in the other Patient ID sequence (0010,1002)** is still not completely resolved, setting `pid.setTypeOfPatientID(TypeOfPatientID.TEXT);` is returning duplicate pids.

Would you suggest how you will resolve that?

Thanks

PR #331, opened 2018-08-06
Little fixes in tools docs.

rrzaripov:


rrzaripov:
@gunterze Thanks for pointing out, I forgot about messages.properties files. Now done.

gunterze:
Thanks too!

PR #340, opened 2018-08-31
Update DICOM native decompressors

nroduit:
Update to opencv 4 (use c++ 11). Update jpeg-ls and openjpeg.
Add native encoders for j2k and jpeg but not activated by default.

Fix decompressing jpeg baseline with YBR_FULL color model:
Example: decompress a jpeg baseline encoded with dcmtk (by default YBR_FULL)
The output color model after decompression must be RGB.


dcm4che-image/src/main/java/org/dcm4che3/image/PhotometricInterpretation.java
```diff
@@ -111,6 +111,11 @@ else if (tsuid.endsWith("91") || tsuid.endsWith("93"))
         public ColorModel createColorModel(int bits, int dataType, Attributes ds) {
             return ColorModelFactory.createYBRFullColorModel(bits, dataType, ds);
         }
+
```
gunterze: Would brake decompression of YBR_FULL RLE (US) images

nroduit: You're right. I forgot this specific case.

However, I think YBR_FULL + RLE should be handled as an exception and all other JPEG decompressions must have an RGB as a color model for the output image.

A much more common case, you compress an image into JPEG Baseline with dcmtk and then decompress it with dcm4chee and you will get an image with a wrong color model (YBR_FULL instead RGB).

dcmtk: dcmcjpeg --encode-baseline color-raw.dcm color-jpeg.dcm
dcm4chee: dcm2dcm color-jpeg.dcm color-raw.dcm

weasis v3 0 3-snapshot_014

gunterze: Will think about to make the Photometric Interpretation of decompressed color images configurable by  ImageWriterFactory.properties.

nroduit: Yes, this will allow having a specific behavior according to the selected decompressor.

nroduit:
I just realized that the Linux build of the native library has been done on Ubuntu 18.04.  So to load this library it requires Linux distribution with GLIBC_2.27. 

It is necessary to rebuild the Linux library to have a backward compatibility with GLIBC_2.23 or GLIBC_2.19.

gunterze:
Would be nice if you could provide binaries linked with GLIBC_2.19 at https://github.com/nroduit/mvn-repo/tree/master/org/weasis/thirdparty/org/opencv/libopencv_java/ 

(Lazy) gunter

PR #350, opened 2018-09-26
Update native codec library to 4.0.0-dcmR2

nroduit:
It should fix the issue of the Linux backward compatibility described [here](https://github.com/dcm4che/dcm4che/pull/340).

Now the backward compatibility of the native library is supported from the most recent version to:
Linux 64-bit: GLIBC_2.14
Linux 32-bit: GLIBC_2.11
Mac OS X: 10.9
Windows 32-bit and 64-bit: at least Windows Vista

This new build has not been tested intensively.


gunterze:
THANKS!

PR #365, opened 2018-11-08
[IEI-73541] 8.1.4 Type 1 attribute TypeOfPatientID(0010,0022) is missing in the otherPatient ID sequence (0010,1002)

mfarooq2000:
@robedwardsagfa please review

Need to set TypeOfPatientID before calling exportPatientIDWithIssuer


robedwardsagfa:
It looks like you're using some changes from master, however you're missing the changes to equals and hashCode in IDWithIssuer.  Was that intentional?

mfarooq2000:
Yes that is correct. If we match object on TypeOfPatientID than we will be
creating duplicate OtherPatientID sequence items for same PatientID with
different Type.

PR #369, opened 2018-11-19
NPE in Association with too many accepted Presentation Contexts

patrick-peer:
The Association code had two flaws:
1) When there were more (or unknown) Presentation Contexts in the Assocate Accept response than in the Associate Request, a NullPointerException was thrown. This, of course, violates the Dicom Standard, but we need to communicate with naughty SCPs, and also were able to do so with dcm4che2 before.
With this patch unknown Presentation Context IDs are ignored in `org.dcm4che3.net.Association.initPCMap()`.
2) Any RuntimeException (i.e. unexpected Exceptions), which also includes the NullPointerException from above, fell out of the Thread in `org.dcm4che3.net.Association.activate()`, omitting any error handling code.
I kind of short circuited this case with
`catch (RuntimeException e) {
    onIOException(new IOException("Unexpected Error", e));
}`
I'm not sure whether this is a smart approach, but it brings the error to the code that tries to open the Association (instead of letting it wait indefinitely). Also this is the least invasive change I could think of, since the error handling is based on IOException, and widening the Exception type here means changes on a larger scale.
I am all ears for better solutions, and also willing to do the work, but for the sake of this pull request I thought less is more.

AssociationTest illustrates the behavior. I encourage you to try it with the old code and then with the fiexes, one by one.


patrick-peer:
So, is there anything I can or should do to bring this forward?

gunterze:
Thanks!

PR #383, opened 2018-12-10
JSONWriter option to write human readable tags

luckfamousa:
Added a constructor to JSONWriter that takes a ElementDictionary as a second parameter.
If this constructor is used the JSONWriter will serialize the Tag names to human readable format using the given ElementDictionary. Can be used like:

`final StringWriter writer = new StringWriter();`
`JsonGenerator gen = Json.createGenerator(writer);`
`new JSONWriter(gen, ElementDictionary.getStandardElementDictionary()).write(attr);`
`gen.flush();`


gunterze:
JSON format is defined by [DICOM PS.3.18, Annex F](http://dicom.nema.org/medical/dicom/current/output/html/part18.html#chapter_F) - see no need to introduce a proprietary flavor

PR #386, opened 2018-12-11
Fix #382 LOG.warn does not properly print integer value on failed VR

### LINKED ISSUE #382: LOG.warn does not properly print integer value on failed VR
**Describe the bug**
LOG.warn has '{0}' string format, which should be '{}' to properly resolve the parameter.
`        LOG.warn("Unrecogniced VR code: {0}H - treat as UN",
                Integer.toHexString(code));`

**To Reproduce**
Still debugging the root cause, so not sure, but causing the exception will throw that log. 
```
 dcm4che/dcm4che-core/src/main/java/org/dcm4che3/data/VR.java
    public static VR valueOf(int code) {
        try {
            VR vr = VALUE_OF[code - AE.code];
            if (vr != null)
                return vr;
        } catch (IndexOutOfBoundsException e) {}
        LOG.warn("Unrecogniced VR code: {0}H - treat as UN",
                Integer.toHexString(code));
        return UN;
}
```

**Expected behavior**
The log should display the code instead of `{0}`

**Screenshots**

**Desktop (please complete the following information):**
Not relevant

**Additional context**
Nothing additional.

hczedik:
Corrected log message. Used Java format (0xXXX) for hex value.

### DIFF
--- a/dcm4che-core/src/main/java/org/dcm4che3/data/VR.java
+++ b/dcm4che-core/src/main/java/org/dcm4che3/data/VR.java
@@ -113,7 +113,7 @@ public enum VR {
             if (vr != null)
                 return vr;
         } catch (IndexOutOfBoundsException e) {}
-        LOG.warn("Unrecogniced VR code: {0}H - treat as UN",
+        LOG.warn("Unrecogniced VR code: 0x{} - treat as UN",
                 Integer.toHexString(code));
         return UN;
     }


gunterze:
Alternatively fixed by  8a9f99b  - prefer ####H over 0x#### in accordance as hex values are labeled in DICOM.

PR #414, opened 2019-03-12
Transcoding

nroduit:
Fix some transcoding issues:
- Conversion of images with a PALETTE color model into jpeg image (the input image is transformed in RGB with 3-byte samples)
- Conversion of raw images with color model YBR_PARTIAL_422, YBR_PARTIAL_420 and YBR_FULL with RLE into jpeg image (the input image is transformed in RGB with 3-byte samples)
- TSUID is changing with files without pixel data (ex. with floating pixel data, the pixels are not compressed but TSUID change to compressed image) => line 350 of Trancoder.java
- Change lossy compression attribute when the compression is not lossless => line 526 of Trancoder.java

Other improvements:
- Add some tests (transcoding YBR_PARTIAL_422 and PALETTE)

Require to apply https://github.com/dcm4che/dcm4che/pull/390


wigun:
same as post in: https://groups.google.com/forum/#!topic/dcm4che/6AMSke72zsk

Unfortunately, before this change the DicomImageReader was more fault tolerant (also like dcm4che2).

For example, if a JPEG Losslesss (1.2.840.10008.1.2.4.70) with encapsulated RGB Data, was incorrectly tagged with Photometric interpretation 'YBR_FULL_422', the DicomImageReader still worked fine and returned the correct BufferedImage.


dcm4che version: 5.17.1:


The statement in line 466:

                if (samples > 1 && bi.getColorModel().getColorSpace().getType() ==
                        (pmiAfterDecompression.isYBR() ? ColorSpace.TYPE_YCbCr : ColorSpace.TYPE_RGB))
                    return bi;


doesn't return the BufferedImage anymore, and in line 494 the construction of the BufferedImage fails due to:

    java.lang.IllegalArgumentException:
    Raster ByteInterleavedRaster: width = 2196 height = 1958 #numDataElements 3 dataOff[0] = 0
    is incompatible with ColorModel ColorModel: #pixelBits = 24 numComponents = 3 color space = org.dcm4che3.image.YBRColorSpace@73e9cf30 transparency = 1 has alpha = false isAlphaPre = false

This is unfavorable, since there are some wrong DICOM files out in the real world :-(

PR #430, opened 2019-04-01
Added quotation marks around JAVA_LIBRARY_PATH in tools executables

Wiles:
Without the quotation marks the command will fail to execute if the path contains spaces.


gunterze:
Does not solve the issue with spaces in $DCM4CHE_HOME

gunterze:
#645

PR #431, opened 2019-04-03
Not enough characters to fill the carry buffer

callumlowcay-rhapsody:
The `characters()` method in `ContentHandlerAdapter` is invoked by a SAX parser when the SAX parser has characters ready for decoding.  The Base64 decoder requires 4 characters at a time, but the SAX parser can send any number of characters.  If the number of characters is not a multiple of 4 then the excess characters are stored in a 'carry buffer'. The next time the `characters()` method is called, we move available characters from the input stream into the carry buffer until there are 4 characters in the carry buffer, then we Base64 decode the carry buffer.

This assumes that there will always be enough characters available to complete the carry buffer.  In the worst case we might need 3 characters, but the SAX parser has no obligation to send us at least 3 characters.  Sometimes it only sends 1 character.  This causes dcm4che to throw an `ArrayIndexOutOfBoundsException` even though the input XML was valid.

The solution is to copy only as many characters as are available to the carry buffer, and if that isn't enough to complete the carry buffer then return and wait for the SAX parser to send some more characters.

### DIFF
--- a/dcm4che-core/src/main/java/org/dcm4che3/io/ContentHandlerAdapter.java
+++ b/dcm4che-core/src/main/java/org/dcm4che3/io/ContentHandlerAdapter.java
@@ -175,6 +175,7 @@ public class ContentHandlerAdapter extends DefaultHandler {
     private void startInlineBinary() {
         processCharacters = true;
         inlineBinary = true;
+        carryLen = 0;
         bout.reset();
     }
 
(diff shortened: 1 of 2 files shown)


hczedik:
Would be good to add a unit test to cover this case.

callumlowcay-rhapsody:
Unit tests added

takch:
Any chance this pull request will be reviewed soon? This change appears to address an issue I have recently encountered.

gunterze:
THANKS!

Issue #450, opened 2019-05-02
Handle wrongly encoded strings more gracefully


hczedik:
**Describe the bug**
We found datasets with wrongly encoded strings that throw an exception on decoding:
```
java.lang.StringIndexOutOfBoundsException: String index out of range: 33
	at java.lang.String.checkBounds(Unknown Source)
	at java.lang.String.<init>(Unknown Source)
	at org.dcm4che3.data.SpecificCharacterSet$Codec.decode(SpecificCharacterSet.java:240)
	at org.dcm4che3.data.SpecificCharacterSet$ISO2022.decode(SpecificCharacterSet.java:467)
	at org.dcm4che3.data.StringValueType.toStrings(StringValueType.java:329)
	at org.dcm4che3.data.VR.toStrings(VR.java:166)
	at org.dcm4che3.data.Attributes.decodeStringValue(Attributes.java:623)
	at org.dcm4che3.data.Attributes.getString(Attributes.java:915)
```

**To Reproduce**
Read a wrongly encoded string.

**Expected behavior**
Decoding errors should be handled more gracefully: Catch them, log them, and then return the default value. (Instead of throwing an exception.)

We will submit a suggested fix in a Pull Request.

PR #457, opened 2019-05-23
Avoid IndexOutOfBounds exception in case of wrongly encoded ISO 2022 string.

ahuser:
Fix dcm4che/dcm4che#450


hczedik:
Hi @gunterze!
@ahuser and me have evaluated some options to handle such wrongly encoded ISO2022 strings more gracefully.
First we wanted to catch exceptions and instead return a default string in Attributes.getString(.) methods. But as there are quite a lot of callers to decode, it is hard to cover all cases.
Now instead we settled on preventing any ArrayIndexOutOfBoundsExceptions within the ISO2022 decode method, and instead try to decode the string as good as we can.
I hope you agree with this approach and can merge this.
Thanks!

gunterze:
merged with minimal changes: https://github.com/dcm4che/dcm4che/commit/2c25025fdda672169f0152bd9117bf50b1f6ff52

PR #510, opened 2019-07-31
Metadata refactor Initial Commit

andrew-cowan:
This is the initial set of changes that I've made to the DCM4CHE3 DicomImageReader.  The goal is to separate the DICOM handling from the image handling by extending the DicomMetaData class.  These changes are not 100% backward compatible with the original implementation of DicomMetaData, so we should discuss whether that will matter for the generic-config branch.

The overview of the changes are available in the following presentation:
[(2) DCM4CHE3 Refactoring.pptx](https://github.com/dcm4che/dcm4che/files/3452356/2.DCM4CHE3.Refactoring.pptx)


dcm4che-core/src/main/java/org/dcm4che3/error/NoPixelDataException.java
```diff
@@ -0,0 +1,31 @@
...
+///////////////////////////////////////////////////////////////
+//
+package org.dcm4che3.error;
+
+import java.io.IOException;
+
+/**
+ * Exception that indicates that the pixel data was missing and could not be returned.
+ * @author Andrew Cowan (awcowan@gmail.com)
+ */
```
andrew-cowan: I've added an explicit indication of a DICOM file that has no pixel data.  This is to disambiguate the cases where we expect pixels but they do not exist in the file.  In DCM4CHE2 we had to figure it out on our own, and it was error prone to properly identify the issue.

dcm4che-core/src/main/java/org/dcm4che3/data/BulkData.java
```diff
@@ -126,6 +132,15 @@ public String getURI() {
...
     }
 
+    public URI toURI() {
+        return URI.create(getURI());
+    }
+
+    public URI toFileURI() {
+        return URI.create(uriWithoutQuery());
+    }
+
```
andrew-cowan: URIs are very important to how BulkData works, so I'm moving toward using typed URIs as much as possible rather than the internal string.  The internal string is used all over the place, so it could not be removed without some major structural changes.

dcm4che-core/src/main/java/org/dcm4che3/data/BulkData.java
```diff
@@ -333,17 +348,11 @@ public String uriWithoutQuery() {
...
             throw new IllegalStateException("uri: null");
- 
-        if (!uri.startsWith("file:"))
-            return new URL(uri).openStream();
-
-        InputStream in = new FileInputStream(getFile());
-        StreamUtils.skipFully(in, offset);
-        return in;
 
+        return this.uriLoader.openStream(this.toFileURI());
```
andrew-cowan: The details of how to open or reuse an image input stream is now assigned to the ImageInputStreamLoader.  This allows deployments to override the IIS behaviour for different types including vendor specific URI schemes.

dcm4che-core/src/main/java/org/dcm4che3/io/stream/CloseableImageInputStreamAdapter.java
```diff
@@ -0,0 +1,79 @@
...
+
+    @Override
+    public long skip(long n) throws IOException {
+        return iis.skipBytes((int) n);
+    }
+
+    @Override
+    public void close() throws IOException {
+        this.iis.close();
+    }
```
andrew-cowan: The existing adapter does not close the underlying stream.  Sometimes you want that behaviour so that we can keep an underlying stream open, but if we know the underlying stream can be reopened from URI, then we want to make sure that we consistently close all underlying streams.

The names aren't great, but I didn't want to change the fundamental design of the adapter in case implementors were relying on streams staying open, or they had already hacked around it.

dcm4che-core/src/main/java/org/dcm4che3/io/stream/DelegateImageInputStream.java
```diff
@@ -0,0 +1,264 @@
...
+import javax.imageio.stream.IIOByteBuffer;
+import javax.imageio.stream.ImageInputStream;
+import java.io.IOException;
+import java.nio.ByteOrder;
+
+/**
+ * Delegate to an underlying ImageInputStream.
+ * @author Andrew Cowan (andrew.cowan@agfa.com)
+ */
+public class DelegateImageInputStream implements ImageInputStream {
```
andrew-cowan: To support the passing of an external IIS into a DicomImageReader, we provide the capability of wrapping that IIS in a delegate that keeps it open.

dcm4che-core/src/main/java/org/dcm4che3/io/stream/ImageInputStreamLoader.java
```diff
@@ -0,0 +1,27 @@
...
+//                                                           //
+///////////////////////////////////////////////////////////////
+//
+package org.dcm4che3.io.stream;
+
+import javax.imageio.stream.ImageInputStream;
+import java.io.IOException;
+
+
+public interface ImageInputStreamLoader<T> {
```
andrew-cowan: Interface to wrap the creation of an ImageInputStream from an input value.  This is required to support delegating of IIS input, but it also allows us to hide the details of service loader binding which can get a bit ugly.

dcm4che-core/src/test/java/org/dcm4che3/io/stream/FileURIImageInputStreamSpiTest.java
```diff
@@ -0,0 +1,49 @@
...
+
+import java.io.File;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+
+import static org.junit.Assert.*;
+
+public class FileURIImageInputStreamSpiTest {
```
andrew-cowan: File URIs are very difficult to work with, and UNC names are the worst.  These tests are designed to catch issues that result from the FileSystem implementation from rejecting invalid URIs.

dcm4che-imageio/src/main/java/org/dcm4che3/imageio/plugins/dcm/DicomAccessor.java
```diff
@@ -0,0 +1,98 @@
...
+
+    /**
+     * Open an ImageInputStream which is partitioned to view only the pixels for the indicated frameIndex.
+     * @param frameIndex 0 based index of the frame to view.
+     * @throws IOException if the pixel data source could not be accessed
+     * @throws IllegalArgumentException if there is no pixel data tag for this object.
+     * @return An open ImageInputStream.
+     */
+    // Do we want to have a PixelDataMissingException that subclasses IOException?
+    ImageInputStream openPixelStream(int frameIndex) throws IOException;
```
andrew-cowan: This is the critical functionality that was extracted from DicomImageReader:  being able to find a frame or consolidate the fragments of an encapsulated single frame image.  ImageInputStream gives us an excellent class to do it with, and it's consistent with the rest of the module.

dcm4che-imageio/src/main/java/org/dcm4che3/imageio/plugins/dcm/DicomAccessor.java
```diff
@@ -0,0 +1,98 @@
...
+
+    /**
+     * Indicate whether pixel data is avilable for this instance.
+     */
+    boolean containsPixelData();
+
+    VR getPixelDataVR();
+
+
```
andrew-cowan: The default methods are provided for ease of use, and allow for access of common DICOM attributes.
It's syntactic sugar, but it does allow implementations to handle potentially difficult situations:  such as poorly written float values, which have caused unexpected parsing errors in the code in the past. 

If people don't like them, then they don't have to use them :)

dcm4che-imageio/src/main/java/org/dcm4che3/imageio/plugins/dcm/DicomMetaData.java
```diff
@@ -38,33 +38,19 @@
...
-public class DicomMetaData extends IIOMetadata {
-
-    private final Attributes fileMetaInformation;
-    private final Attributes attributes;
-
-    public DicomMetaData(Attributes fileMetaInformation, Attributes attributes) {
-        this.fileMetaInformation = fileMetaInformation;
-        this.attributes = attributes;
-    }
+public abstract class DicomMetaData extends IIOMetadata implements DicomAccessor {
```
andrew-cowan: This is the main area of incompatibility with the current generic-config branch:  DicomMetadata is now an abstract class.  I don't think very many people are instantiating DicomMetaData directly, I believe the DicomMetaData class is generally created by DicomImageReader and then accessed from getMetaData() method, so I don't think that's a serious issue for most users.

dcm4che-imageio/src/main/java/org/dcm4che3/imageio/plugins/dcm/ParseEntireFileMetaData.java
```diff
@@ -0,0 +1,191 @@
...
+import java.net.URL;
+import java.nio.ByteOrder;
+import java.nio.file.Path;
+
+/**
+ * Load the entire file into memory as BulkData.  Pixels are streamed from the BulkData URI to clients.  This class
+ * sacrifices initial parse time for having a stateless class which is able to access the same file from different threads
+ * without synchronization.
+ *
+ */
```
andrew-cowan: The initial implementation just reads the whole file.  I'm planning on extracting the code that was necessary for reading incrementally through the pixel data in a new DicomMetaData implementation in a subsequent release.  My goal is to enable implementations to customize the metadata handling to efficiently handle some of the very large instances that are being generated by modalities today.  

Reading the entire file and reading to the pixel data are two obvious implementations, but i'm hoping that vendors find new ways to override the metadata reading in the future.  It's not even necessary to start with Part 10 files:  we could conceivably write a Metadata implementation that is based on retrieval of JSON from webservices...

dcm4che-imageio/src/main/java/org/dcm4che3/imageio/stream/BulkURIImageInputStream.java
```diff
@@ -0,0 +1,84 @@
...
+import javax.imageio.stream.ImageInputStream;
+import javax.imageio.stream.ImageInputStreamImpl;
+import java.io.IOException;
+
+/**
+ * Return a subset of an image input stream.  The local variables track the location within the offset/length
+ * bounds which are mapped onto the underlying full image input stream.
+ * @author Andrew Cowan (awcowan@gmail.com)
+ */
+public class BulkURIImageInputStream extends ImageInputStreamImpl {
```
andrew-cowan: Really simple segmented IIS implementation.

The current SegmentedImageInputStream has a lot of knowledge about how BulkData fragments are stored, which is fine, but I would like to be able to generically point to a subset of a URI and this class provides that.

dcm4che-imageio/src/main/java/org/dcm4che3/imageio/stream/ByteArrayImageInputStream.java
```diff
@@ -0,0 +1,51 @@
...
+//
+package org.dcm4che3.imageio.stream;
+
+import javax.imageio.stream.ImageInputStreamImpl;
+import java.io.IOException;
+
+/**
+ * ImageInputStream which wraps an in-memory byte[]
+ * @author Andrew Cowan (andrew.cowan@agfa.com)
+ */
```
andrew-cowan: Efficient IIS implementation for byte[].  You can do the same thing with an InputStream, but this will allow random access to the byte[] that is already in memory:  if we already have the data in memory, we might as well use it.

dcm4che-imageio/src/main/java/org/dcm4che3/imageio/stream/ImageInputStreamAdapter.java
```diff
@@ -38,62 +38,26 @@
...
  *
  */
-public class ImageInputStreamAdapter extends InputStream {
-
-    private final ImageInputStream iis;
-
-    private long markedPos;
-
-    private IOException markException;
+public class ImageInputStreamAdapter extends CloseableImageInputStreamAdapter {
```
andrew-cowan: Saving some code by subclassing and suppressing close()

dcm4che-imageio/src/main/java/org/dcm4che3/imageio/stream/SegmentedImageInputStream.java
```diff
@@ -257,4 +257,22 @@ public long getLastSegmentEnd() {
             return bulk.getSegmentEnd();
         }
     }
+
+    /**
+     * Determine the length of the total segmented stream.
+     */
```
andrew-cowan: It's *VERY* useful to know the length of a stream, so I have added length() into all of the IIS implementations.

dcm4che-all/pom.xml
```diff
@@ -42,7 +42,7 @@
   <parent>
     <artifactId>dcm4che-parent</artifactId>
     <groupId>org.dcm4che</groupId>
-    <version>3.3.9-SNAPSHOT</version>
+    <version>3.3.9-METADATA-SNAPSHOT</version>
```
andrew-cowan: necessary evil of using maven...

dcm4che-core/src/main/java/org/dcm4che3/error/NoPixelDataException.java
wayfarer3130: Do you want to allow distinguishing between no pixel data at all and no pixel data for a given frame, at least in the message?  It seems like that is an interesting thing to know when things go wrong.

dcm4che-core/src/main/java/org/dcm4che3/io/stream/DelegateImageInputStream.java
```diff
@@ -0,0 +1,264 @@
...
+        return iis.isCachedMemory();
+    }
+
+    @Override
+    public boolean isCachedFile() {
+        return iis.isCachedFile();
+    }
+
+    @Override
+    public void close() throws IOException {
```
wayfarer3130: Should you set the iis to null here so that it is immediately known that this is closed if you try to re-use it?  I'm just concerend about someone using one of these and not realizing that it wasn't being used correctly.

dcm4che-core/src/main/java/org/dcm4che3/util/StreamUtils.java
```diff
@@ -150,8 +153,7 @@ public static void copy(InputStream in, OutputStream out, int len,
         }
     }
 
-    public static void copy(InputStream in, OutputStream out, int len,
-            int swapBytes) throws IOException {
+    public static void copy(InputStream in, OutputStream out, int len, int swapBytes) throws IOException {
```
wayfarer3130: If you are going to tweak the signature, could you make that len a long instead?  Or perhaps automatically fix it if it is a -ve value?

dcm4che-imageio/src/main/java/org/dcm4che3/imageio/plugins/dcm/DicomImageReader.java
```diff
@@ -181,113 +131,27 @@
...
-    private void initPixelDataFile() {
-        if (pixelData != null)
-            pixelDataFile = pixelData.getFile();
-        else if (pixelDataFragments != null && pixelDataFragments.size() > 1) {
-            Object frag = pixelDataFragments.get(1);
-            if( frag instanceof BulkData ) {
-                pixelDataFile = ((BulkData) frag).getFile();
-            }
+        if (input instanceof DicomMetaData) {
+            setMetadata((DicomMetaData) input);
```
wayfarer3130: I like this setup to handle input being a dicom meta data - that feels like a reasonable way to work with this.

dcm4che-imageio/src/main/java/org/dcm4che3/imageio/plugins/dcm/DicomImageReader.java
```diff
@@ -394,63 +236,64 @@ public boolean canReadRaster() {
...
     public Raster readRaster(int frameIndex, ImageReadParam param)
             throws IOException {
         readMetadata();
         checkIndex(frameIndex);
 
-        openiis();
-        try {
+        // Need to map from Fragment to IIS.
+        Raster raster;
+        try (ImageInputStream iisOfFrame = iisOfFrame( frameIndex)) {
```
wayfarer3130: This is MUCH cleaner doing the open/close around this and not worrying about the underlying state.

dcm4che-imageio/src/main/java/org/dcm4che3/imageio/plugins/dcm/ParseEntireFileMetaData.java
```diff
@@ -0,0 +1,191 @@
...
+
+    public ParseEntireFileMetaData(Object input) throws IOException {
+        this.uri = toURI(input);
+
+
+        String uriStr = this.uri.toString();
+        ImageInputStream iis = SERVICE_LOADER.openStream(input);
+        try (DicomInputStream dis = new DicomInputStream(new ImageInputStreamAdapter(iis))) {
+            dis.setIncludeBulkData(DicomInputStream.IncludeBulkData.URI);
+            dis.setBulkDataDescriptor(BulkDataDescriptor.PIXELDATA);
```
wayfarer3130: It should be possible to provide the preferred bulk data descriptor, as PIXELDATA isn't really an optimal choice for large private tags data.

dcm4che-imageio/src/main/java/org/dcm4che3/imageio/plugins/dcm/ParseEntireFileMetaData.java
```diff
@@ -0,0 +1,191 @@
...
+            dis.setURI(uriStr);
+
+            this.fmi = dis.getFileMetaInformation();
+            this.attributes = dis.readDataset(-1,-1);
+            this.pixelDataVR = attributes.getVR(Tag.PixelData);
+            this.pixelData = attributes.remove(Tag.PixelData);
```
wayfarer3130: I very much like the fact that by the time this method is done, the pixelData is fully defined, although I suspect it will be necessary to have various lazy versions of this that use a separate thread to finish filling out the data.

dcm4che-core/src/main/java/org/dcm4che3/io/stream/DelegateInputStreamSpi.java
```diff
@@ -0,0 +1,52 @@
...
+import javax.imageio.stream.ImageInputStream;
+import java.io.File;
+import java.io.IOException;
+import java.util.Locale;
+
+/**
+ * Wrap an existing image input stream and allow it to be used without closing the underlying input stream.
+ * @author Andrew Cowan (andrew.cowan@agfa.com)
+ */
+public class DelegateInputStreamSpi extends ImageInputStreamSpi {
```
robedwardsagfa: Minor nitpick, can this be DelegateImageInputStreamSpi to stay consistent with your other naming?

dcm4che-imageio/src/main/java/org/dcm4che3/imageio/plugins/dcm/DicomMetaData.java
robedwardsagfa: Is it ever used in our code base directly?  I know the goal of some is to sync generic-config up with trunk, however realistically I don't see that happening, so this might be ok.

dcm4che-core/src/main/java/org/dcm4che3/data/BulkData.java
```diff
@@ -333,17 +348,11 @@ public String uriWithoutQuery() {
         return uri.substring(0, uriPathEnd);
     }
 
-    public InputStream openStream() throws IOException {
+    public ImageInputStream openImageInputStream() throws IOException {
```
hczedik: As you removed the openStream() method (replaced it with the openImageInputStream()), does that mean the BulkData now only works for pixeldata, and no longer for other BulkData like overlay data etc.?

dcm4che-core/src/main/java/org/dcm4che3/io/stream/CloseableImageInputStreamAdapter.java
```diff
@@ -0,0 +1,79 @@
...
+//
+package org.dcm4che3.io.stream;
+
+import javax.imageio.stream.ImageInputStream;
+import java.io.FilterInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+
+/**
+ *
```
hczedik: Would be good to add some Javadoc with a link to the existing adapter and explaining the differences.

dcm4che-core/src/main/java/org/dcm4che3/error/NoPixelDataException.java
andrew-cowan: Yes.  I have extended the errors that are reported in the latest build:

- NoPixelDataException: Exception that indicates that the pixel data was missing and could not be returned.
- TruncatedPixelDataException: Thrown when there are fewer bytes available in the stream than expected,
- UnavailableFrameException: Thrown when a frame is requested that does not exist in the underlying instance.

dcm4che-imageio/src/main/java/org/dcm4che3/imageio/plugins/dcm/DicomMetaData.java
andrew-cowan: It turns out that the constructor WAS being used in the codebase:  the preprocessing code was creating it's own DicomMetadata object and passing it into DIcomImageReader manually.  If they were doing it, then likely other people will too.

I modified the code to restore the old constructor with the old functionality.  I should not have taken the risk of breaking that backward compatibility in the first place :(

dcm4che-core/src/main/java/org/dcm4che3/io/stream/DelegateImageInputStream.java
andrew-cowan: mmmm, it's worth being safe than sorry here.  If I add an accessor for iis then I can do that with minimal impact to the class.  I'll do that in my local branch.

andrew-cowan: I updated the code to throw an IllegalStateException if the delegate is used after the close() is invoked.  That's not the best exception to throw, but i needed something that was Runtime so it could be thrown by all methods, and I didn't want to add a new exception for something this small...

dcm4che-core/src/main/java/org/dcm4che3/io/stream/DelegateInputStreamSpi.java
andrew-cowan: Mmm, good point.  Fixed.

dcm4che-core/src/main/java/org/dcm4che3/util/StreamUtils.java
andrew-cowan: I have changed the code to pass long values.  I'm going to run some tests to make sure that it works properly.

dcm4che-imageio/src/main/java/org/dcm4che3/imageio/plugins/dcm/ParseEntireFileMetaData.java
andrew-cowan: The original implementation relied on people doing DicomMetaData subclasses to modify how the data was read.  I have addressed that in a number of ways i the latest build:  

- DicomMetaData is a concrete class again, and it defers the reading of attributes / values to DicomImageAccessor which is passed in the constructor.   This means that the Accessor can get data in any manner the application wants.
- If the user has the Attributes / FMI then they can be read in the way they want and passed through the constructor.
- The DefaultMetaDataFactory is a configurable factory which takes a URI and generates an appropriate DicomImageAccessor subclass.  One of the configurations on the factory is BulkDataDescriptor.

dcm4che-imageio/src/main/java/org/dcm4che3/imageio/plugins/dcm/ParseEntireFileMetaData.java
andrew-cowan: Yes, I think so too.  For the default implementation  I wanted something very simple.  The DicomImageAccessor can accommodate more complex designs, as the openPixelStream(int frameIndex) call hides the particulars.

dcm4che-core/src/main/java/org/dcm4che3/data/BulkData.java
andrew-cowan: No, it works for all types of BulkData:  I'm just using IIS internally because it allows positional changes and skipping.  In a subsequent update I restored the openStream() which wraps the underlying IIS.

dcm4che-core/src/main/java/org/dcm4che3/io/stream/CloseableImageInputStreamAdapter.java
andrew-cowan: Good point:  I've actually forgetten the differences myself now :-/

andrew-cowan: I added comments to DelegateImageInputStream and CloseableImageInputStreamAdapter.  I deprecated the ImageInputStreamAdapter as the same behaviour can be created using the other two adapters.

dcm4che-core/src/main/java/org/dcm4che3/util/StreamUtils.java
andrew-cowan: I have made the change locally so that a long is being passed and cast down to int internally, but I am concerned about side-effects.  There is no explicit StreamUtils coverage in DCM4CHE3 right now, and I think we should make a separate issue for handling that.  

I'm not really happy with StreamUtils in general to be honest, and we should fix it to handle not just large objects, but also to more stringently handle the InputStream being passed.  The code is consistently using read() == 0 as the end of file, which is not correct.  the InputStream is complete when read() == -1.

The main advantage this class has over the commons IOUtil is that it supports byte swapping.  Perhaps if we added an InputStream or OutputStream implementation that would do swapping inline we could just use the commons IO and really simplify this code...

dcm4che-imageio/src/main/java/org/dcm4che3/imageio/plugins/dcm/DicomImageReader.java
andrew-cowan: Thanks.  A lot of my changes are around making DicomImageReader more flexible, and providing client applications more options in using it.

dcm4che-core/src/main/java/org/dcm4che3/util/StreamUtils.java
andrew-cowan: I actually think the best approach would be to write new methods which take long values and use IIS directly.  So for instance:

copy(ImageInputStream, OutputStream): long
copy(ImageInputStream, ImageOutputStream): long
copy(ImageInputStream, WritableChannel):long
copy(ImageInputStream, OutputStream, long offset, long length) 

etc...

The code is using IIS for all I/O, we should use the interface directly so that we get the most benefits.  We can either add those methods to StreamUtils directly, or even add a new StreamHandler interface that might allow us to intercept how copies are being done.  That would give us a nice way to override the buffer sizes or even add thread local buffering if required by an application.

dcm4che-imageio/src/main/java/org/dcm4che3/imageio/plugins/dcm/DicomImageReader.java
andrew-cowan: Thanks.  It is also friendly for IDEs which like to detect resource leaks.

dcm4che-core/src/main/java/org/dcm4che3/data/Fragments.java
```diff
@@ -256,4 +256,21 @@ private int itemHashCode(Object e) {
         }
     }
 
+
+
+    /**
+     * Determine the length of an object found in the BulkData
```
hczedik: You mean the length of an item in the Fragments? For consistency with the other methods it should be called `itemLength(Object e)`

dcm4che-core/src/main/java/org/dcm4che3/io/BulkDataDescriptor.java
```diff
@@ -102,6 +113,29 @@ public boolean isBulkData(List<ItemPointer> itemPointer, String privateCreator,
         };
     }
 
+    public static BulkDataDescriptor or(BulkDataDescriptor... descriptors) {
+        return new BulkDataDescriptor() {
+            @Override
+            public boolean isBulkData(List<ItemPointer> itemPointer, String privateCreator, int tag, VR vr, int length) {
+                for(BulkDataDescriptor bd : descriptors) {
+                    if(bd.isBulkData(itemPointer, privateCreator, tag,vr,length))
```
hczedik: Formatting: a few spaces missing here (between method arguments, and directly after "if" and for").

dcm4che-core/src/main/java/org/dcm4che3/io/stream/DelegateImageInputStream.java
```diff
@@ -0,0 +1,276 @@
...
+    }
+
+    @Override
+    public boolean isCachedFile() {
+        return iis().isCachedFile();
+    }
+
+    @Override
+    public void close() throws IOException {
+        // Destroy the delegate
```
hczedik: The comment "Destroy the delegate" is a bit misleading. It will certainly not be destroyed by settings the reference to null. I would rather replace it with something like: "Ensure that delegate cannot be accessed any more".

hczedik: Okay, I think I read it wrong: What confused me is that you refer to this class as "the delegate". I was assuming you mean the field "iis" with "the delegate". (Which I think makes more sense.)

dcm4che-core/src/main/java/org/dcm4che3/io/stream/DelegateImageInputStream.java
hczedik: The name is confusing. For me a "delegate" is the inner field ("iis") that you are delegating to. So the wrapping class should not also be called "delegate". Maybe better: "DelegatingImageInputStream"? Or "NotCloseableImageInputStreamWrapper"?

dcm4che-core/src/main/java/org/dcm4che3/io/stream/DelegateImageInputStream.java
```diff
@@ -0,0 +1,276 @@
...
+
+    @Override
+    public String toString() {
+        return "Delegate("+iis+")";
+    }
+
+
+    protected ImageInputStream iis() throws IllegalStateException {
+        if(iis == null) {
+            throw new IllegalStateException("Delegate is already closed.");
```
hczedik: It's not the "delegate" that is already closed, but this class (which in my opinion should not be called "delegate").

dcm4che-imageio/src/main/java/org/dcm4che3/imageio/codec/ImageReaderFactory.java
```diff
@@ -73,9 +73,11 @@
 @LDAP(objectClasses = "dcmImageReaderFactory")
 @ConfigurableClass
 public class ImageReaderFactory implements Serializable {
-
     private static final Logger LOG = LoggerFactory.getLogger(ImageReaderFactory.class);
 
+    /** Optional flag that */
```
hczedik: "Optional flag that..." Did you want to finish that sentence? :-)

dcm4che-imageio/src/main/java/org/dcm4che3/imageio/stream/ImageInputStreamAdapter.java
```diff
@@ -38,62 +38,28 @@
...
 import java.io.InputStream;
 
 import javax.imageio.stream.ImageInputStream;
 
 /**
+ * ImageInputStream which does not close the underlying stream.
  * @author Gunter Zeilinger <gunterze@gmail.com>
- *
+ * @deprecated ImageInputStreams should always be closed.  If an independant delegate is required then wrap a
+ * DelegateImageInputStream with CLoseableImageInputStreamAdapter.
```
hczedik: Typo in class name: CLoseableImageInputStreamAdapter
Would be good to do actual links to the classes with {@link ...}.

hczedik:
@andrew-cowan: I reviewed everything now. Very nice work! I know my review is a bit too late, but I hope you can consider some of my remarks in a future commit/merge. 
I also hope that you can start now the work to port these changes over to master. I also hope that @gunterze will accept them then. It would make sense though to split it into a few iterative steps (Pull Requests), to ease the review process for him.

PR #525, opened 2019-08-28
Generic config

MarkDechamps:
Don't allow duplicate ids. So if the root id already 'matches' one of the ids in otherPatientIdSequence, then don't add it again.


MarkDechamps:
Don't allow duplicate ids. So if the root id already 'matches' one of the ids in otherPatientIdSequence, then don't add it again.

hczedik:
We should adapt this pull request to match what Gunter has committed to the master branch: See #529

hczedik:
Closing this PR as it has not seen any activity since 2019 and an improved version has been committed in https://github.com/dcm4che/dcm4che/issues/529

PR #526, opened 2019-08-28
Only add root id to OtherPatientIdSequence if it doesn't match with any id in there already.

MarkDechamps:
Only add root id to otherPatientIds if it doesn't already match with any id in there.Otherwise we can get duplicate ids.


dcm4che-core/src/test/java/org/dcm4che3/data/IDWithIssuerUnitTest.java
```diff
@@ -0,0 +1,84 @@
+package org.dcm4che3.data;
```
hczedik: License block (see other files) should be added also here.

dcm4che-core/src/test/java/org/dcm4che3/data/IDWithIssuerUnitTest.java
```diff
@@ -0,0 +1,84 @@
...
+
+import org.junit.Test;
+
+import java.util.Set;
+
+import static org.dcm4che3.data.Tag.*;
+import static org.junit.Assert.assertEquals;
+
+public class IDWithIssuerUnitTest {
```
hczedik: To be consistent with the other names in this package, the unit test class should be called "IDWithIssuerTest".

Please also add a class Javadoc with author (and maybe short explanaition and link to the class under test).

dcm4che-core/src/main/java/org/dcm4che3/data/IDWithIssuer.java
```diff
@@ -149,16 +149,19 @@ public boolean equals(Object obj) {
                     ? other.typeOfPatientID == null
                     : typeOfPatientID.equals(typeOfPatientID)) &&
                 (identifierTypeCode == null
-                    ? other.identifierTypeCode == null
-                    : identifierTypeCode.equals(identifierTypeCode)) &&
+                        ? other.identifierTypeCode == null
```
hczedik: Please remove additional formatting.

gunterze:
Resolved by more generic https://github.com/dcm4che/dcm4che/issues/529

PR #535, opened 2019-09-07
dcm4che#532 fix timeouts for outgoing requests

mbakhoff:
Before commit 6b2dbf6 the response timeout applied to both sending out a
request and waiting for a response for it. After the change it is impossible
to set a timeout for sending the request.


gunterze:
fixed by https://github.com/dcm4che/dcm4che/commit/17f9d5fb60120845bdb88e2bc1d9fe69210effff

PR #549, opened 2019-09-24
[IEI-96376] Fix CStoreSCUResp when extending response

homerocda:
Error information was lost in IEI-96376 because BasicCStoreSCUResp removes the Last Error when the addendum response doesn't have any errors. The correct behavior should be replacing it only if a new one arises.


homerocda:
Please note that there are breaking changes to EI now. Pulling this needs to be coordinated with the appropriate changes in pacs-core-server.

Issue #552, opened 2019-09-30
DCM4CHE doesn't keep block identification code for private elements


homerocda:
**Describe the bug**
We have an issue where the server may change data attribute tags when importing objects. One element with tag (3773,2030) was changed to (3773,1030). In another file, (3773,3025) was changed to (3773,1025). The culprit was found to be the way that DCM4CHE handles private tags.

**To Reproduce**
This test fails:

       Attributes original = new Attributes();
        original.setString("PrivateCreatorA", 0x00991001, VR.LO, "0099xx01A");
        original.setString("PrivateCreatorB", 0x00992001, VR.LO, "0099xx01B");
        original.setString("PrivateCreatorC", 0x00993001, VR.LO, "0099xx01C");

        Attributes selection = new Attributes();
        selection.setNull("PrivateCreatorA", 0x00991001, VR.LO);
        selection.setNull("PrivateCreatorC", 0x00993001, VR.LO);

        Attributes filtered = new Attributes();
        filtered.addSelected(original, selection);

        assertTrue("Attributes should keep the element tag for PrivateCreatorA", filtered.contains(0x00991001)); // PASSES
        assertTrue("Attributes should keep the element tag for PrivateCreatorC", filtered.contains(0x00993001)); // FAILS!

**Expected behavior**
Test should pass and filtered should contain the correct tags.

**Desktop (please complete the following information):**
 - OS: Windows 10

**Additional context**
When parsing a DICOM object, private creator elements are ignored at first. Only when the first private data element is received it will insert the private creator tag. It does so by taking the group number from the data element tag, appending a 0x10 suffix to it and checking for conflicts.   If a conflict is found, the tag is incremented until a free private element block is found. The private element block number is completely ignored in this process. Refer to (https://github.com/dcm4che/dcm4che/blob/abd3179558166dc8e03d1a897352efb96f46a9d5/dcm4che-core/src/main/java/org/dcm4che3/data/Attributes.java#L406)

hczedik:
Might be related to https://dcm4che.atlassian.net/browse/LIB-362

homerocda:
Hi @hczedik. #40 is related to this. We are using the generic-config branch.

gunterze:
Fixed by #554

Issue #40, opened 2017-03-15
Modify Attributes.addAll() to preserve non-conflicting Private Creator ID tag positions


hczedik:
Attributes.addAll() should preserve non-conflicting Private Creator ID tag positions.

This behavior was originally implemented in commit https://github.com/dcm4che/dcm4che/commit/7c1d8f4900f2c89a79efc223ab9e672b199da1f9

Unfortunately it was broken with the bug fix for http://dcm4che.org/jira/browse/LIB-362 (Commit https://github.com/dcm4che/dcm4che/commit/764d9c4f25da1d87f119c057fcf76227771e3a12 ).

The behavior has to be restored (of course also preserving the bug fix for LIB-362).

New unit tests (AttributesTest) should be implemented to ensure that this behavior doesn't break again.

hczedik:
Actually preserving the positions might not always be possible, as the standard says that the "first unused" block should be used.
See part 5 7.8.1 Private Data Element Tags: The implementer shall insert an identification code in the first unused (unassigned) Element in this series to reserve a block of Private Elements.

hczedik:
Fixed by https://github.com/dcm4che/dcm4che/pull/554

PR #554, opened 2019-10-02
[LIB-362] Wrong private tag handling in Attributes.addSelected()

hczedik:
Cherry-pick of 764d9c4f25da1d87f119c057fcf76227771e3a12 which was initially submitted as pull request #21 but never ported over to the new master branch.

See https://dcm4che.atlassian.net/browse/LIB-362

@gunterze This is an old fix from the generic-config branch. It fixes LIB-362 but causes #40. So I am not sure if you want to accept this.
(See also the discussion in the new issue #552.)
What do you think?

### DIFF
--- a/dcm4che-core/src/main/java/org/dcm4che3/data/Attributes.java
+++ b/dcm4che-core/src/main/java/org/dcm4che3/data/Attributes.java
@@ -2159,6 +2159,11 @@ public class Attributes implements Serializable {
                 }
             }
             if (!simulate) {
+                if (privateCreator != null
+                        && creatorTagOf(privateCreator, tag, false) < 0
+                        && !contains(creatorTag)) {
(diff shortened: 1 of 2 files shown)


hczedik:
Actually I think the fix is fine as it is, as the standard says that always the "first unused" block should be used. (See part 5 7.8.1 Private Data Element Tags: The implementer shall insert an identification code in the first unused (unassigned) Element in this series to reserve a block of Private Elements.)

The current code can cause gaps between the reserved blocks, the change ensures that always the first unused block gets reserved.
Do you agree @gunterze ?

gunterze:
Want to finish another 2-3 day task, before reviewing the proposed change.

Sent with [ProtonMail](https://protonmail.com) Secure Email.

‐‐‐‐‐‐‐ Original Message ‐‐‐‐‐‐‐

gunterze:
Should be fixed by https://github.com/dcm4che/dcm4che/commit/85363bf860639ed94c7c737790964e1c51556ce6

PR #578, opened 2019-11-04
[IEI-87704] AuditLogger.isInstalled assumes null = true for Boolean v…

qzhang0118:
When an AuditLogger extension exists in the configuration if the 'dicom installed' value has never been set or modified it will be null in the actual configuration.

This will result in audit messages still being created and put into memory even if there is no place for them to be sent or saved to with auditing disabled completely.

### DIFF
--- a/dcm4che-net-audit/src/main/java/org/dcm4che3/net/audit/AuditLogger.java
+++ b/dcm4che-net-audit/src/main/java/org/dcm4che3/net/audit/AuditLogger.java
@@ -334,11 +334,11 @@ public class AuditLogger extends DeviceExtension {
     @ConfigurableProperty(name = "dcmAuditMessageSupplement95Schema", defaultValue = "false")
     private boolean supplement95;
 
-    @ConfigurableProperty(name = "dicomInstalled")
-    private Boolean auditLoggerInstalled;
+    @ConfigurableProperty(name = "dicomInstalled", defaultValue = "false")
+    private boolean auditLoggerInstalled = false;
 
     @ConfigurableProperty(name = "dcmAuditIncludeInstanceUID")
-    private Boolean doIncludeInstanceUID = false;
+    private boolean doIncludeInstanceUID = false;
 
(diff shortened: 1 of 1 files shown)


dcm4che-net-audit/src/main/java/org/dcm4che3/net/audit/AuditLogger.java
```diff
@@ -637,30 +637,29 @@ public void setSupplement95(boolean sup95) {
 
     public boolean isInstalled() {
         return device != null && device.isInstalled()
-                && (auditLoggerInstalled == null || auditLoggerInstalled.booleanValue());
+                && auditLoggerInstalled;
```
homerocda: Won't this break the current behavior? Are there any use cases where AuditLoggerInstalled can be null?

qzhang0118:
Still need to change the version of dcm4che3 in the following components:
agility-server-common
pacs-core-server
asb

dcm4che-net-audit/src/main/java/org/dcm4che3/net/audit/AuditLogger.java
qzhang0118: I think the auditLoggerInstalled is null only when it's first set to it as the default value. I don't see why it needs to be null otherwise.
The old logic is causing audit messages being generated even if auditing is disabled.

gunterze: No. Should behave like the Installed field of [Network Application Entity](http://dicom.nema.org/medical/dicom/current/output/html/part15.html#sect_H.1.1.2) and [Network Connection](http://dicom.nema.org/medical/dicom/current/output/html/part15.html#sect_H.1.1.3):

> A Boolean value. True if the Audit Logger is installed. If null, information about the installed status of the Audit Logger  is inherited from the device.

gunterze:
Will **not** be merged to master!

PR #586, opened 2019-11-15
IEI-81693 - Allow implicit relational behavior for CFIND/CMOVE

homerocda:
Changes to allow an AE to accept relational CFIND/CMOVE requests without requiring it to be negotiated during the association.

### DIFF
--- a/dcm4che-net/src/main/java/org/dcm4che3/net/Status.java
+++ b/dcm4che-net/src/main/java/org/dcm4che3/net/Status.java
@@ -275,12 +275,6 @@ public final class Status {
     public static final int ElementsDiscarded                  = 0xB006;
     public static final int DataSetDoesNotMatchSOPClassWarning = 0xB007;
 
-    /**
-     * Warning: Relational Behavior Not Negotiated (B1xxH): relational behavior
-     * was allowed but not negotiated during association.
-     */
-    public static final int RelationalBehaviorNotNegotiated    = 0xB100;
-
     public static final int UnableToProcess                    = 0xC000;
     public static final int CannotUnderstand                   = 0xC000;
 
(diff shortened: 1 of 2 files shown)


dcm4che-net/src/main/java/org/dcm4che3/net/Status.java
```diff
@@ -275,11 +275,20 @@
     public static final int ElementsDiscarded                  = 0xB006;
     public static final int DataSetDoesNotMatchSOPClassWarning = 0xB007;
 
+    /**
+     * Warning: Relational Behavior Not Negotiated (B1xxH): relational behavior
+     * was allowed but not negotiated during association.
+     */
+    public static final int RelationalBehaviorNotNegotiated    = 0xB100;
```
hczedik: Why 0xB100? That is not defined in the standard, right?

homerocda: The standard defines BXXX as a warning indication. I'm setting B1 so we can still have the last 2 bytes for finer warning indications.

gunterze: DICOM does not permit any Bxxxx Warning Status on [any C-FIND service](http://dicom.nema.org/medical/dicom/current/output/html/part07.html#sect_9.1.2.1.6), and only B000H (Sub-operations Complete - One or more Failures) for [C-MOVE operations of the Retrieve Service Classes](http://dicom.nema.org/medical/dicom/current/output/html/part04.html#sect_C.4.2.1.5).

gunterze: s.o. https://github.com/dcm4che/dcm4che/issues/587

homerocda: This status was removed in the latest commit, and we will return Success instead. This commit still should go into generic-config to enable the implicit relational behavior.

PR #590, opened 2019-11-29
DicomInputStream: change some fields from private to protected

craigtmoore:
Update DicomInputStream so that I can extend it in a external class.


dcm4che-core/src/main/java/org/dcm4che3/io/DicomInputStream.java
```diff
@@ -107,12 +107,12 @@
...
     private String uri;
-    private String tsuid;
+    protected String tsuid;
     private byte[] preamble;
     private Attributes fileMetaInformation;
     private boolean hasfmi;
-    private boolean bigEndian;
-    private boolean explicitVR;
+    protected boolean bigEndian;
+    protected boolean explicitVR;
```
homerocda: Wouldn't it be better to add a public or protected setter instead of changing these field's access level?

craigtmoore: This is not meant to be a Java enterprise Bean (with get/set). These fields are only used internally so I think its acceptable to access them this way.

PR #603, opened 2019-12-18
Correct name of b132 buffer and add some documentation to guessTransferSyntax

hczedik:
@gunterze: Just a small update to documentation based on our recent communication.


gunterze:
Why to rename b132 to b128? The naming just reflects the length of the byte array. The reason for choosing 132 and not 128 is to read the 128 byte File Preamble and following 4 byte DICOM prefix by one read operation.

hczedik:
But I did correct the name from b128 to b132.

hczedik:
Ups, but I did make a mistake in the documentation... I wrote 132-8=128. I'll correct that.

gunterze:
Thx!

hczedik:
Documentation is now corrected.

Issue #599, opened 2019-12-11
Optionally include unused bits of pixel value in compressed stream


gunterze:
Configurable by _Image Write Param_ `bitsCompressed=16`

_Rationale_: https://github.com/dcm4che/dcm4chee-arc-light/issues/2281

PR #616, opened 2020-01-12
Fix short data issues 599

nroduit:
The issue https://github.com/dcm4che/dcm4chee-arc-light/issues/2281 comes from the nullify process (for removing embedded overlay) which does not handle correctly signed 12-bit dataBuffer.

JPEG Lossless must be encoded as signed data [according to the standard](http://dicom.nema.org/medical/dicom/current/output/chtml/part05/sect_8.2.html#table_8.2.1-2) and the native codec requires shortDataBuffer for signed data.

And the last commit fixes reading less than 16-bit (Bitstored) short data.


nroduit:
This fix should make it possible to go back on this commit https://github.com/dcm4che/dcm4chee-arc-light/commit/44a4f69ae94c8e495c75032bad37c268e67ab735 and not use by default BitStored at 16 https://github.com/dcm4che/dcm4che/issues/599

gunterze:
I prefer to stay with including unused bits in the compression stream by default configuration. With  #638, it's limited to signed pixel data. Compressing the images attached to dcm4che/dcm4chee-arc-light#2281 JPEG Lossless with 12 vs. 16 bits per sample shows an  increase of the compressed size of only 1.2 - 1.5%.   Caused by #637 , I could not compare the compression size for JPEG LS and J2K.

Would rather provide functionality for sign extension for signed short pixel values independently from decompression.

nroduit:
For signed values, all encoders expect to receive a short data buffer (with values corresponding to 16-bit signed representation - Two’s Complement). In short, it is sufficient to transmit signed data with a 16-bit representation to all encoders (for 12-bit JPEG, the 12-bit representation is handled by the encoder).

So changing the output bitstored to 16-bit will not fix all the issues. If you try to convert less than 16-bit signed pixels, the output data will be corrupted. Here some samples:
http://www.dclunie.com/images/signedrange/index.html
ftp://dicom.offis.de/pub/dicom/offis/software/ihe_y2/release_2004/imgdisplay_testcases.zip
=> Modality LUT and VOI LUT

The problem comes when reading uncompressed signed data with a bitstored less 16-bit. The representation of the pixel value is related to bitstored, for instance the decimal value -55 at different bitstored:
```
10 => 0000001111001001
12 => 0000111111001001
16 => 1111111111001001
```
So it was the purpose of my commit https://github.com/dcm4che/dcm4che/pull/616/commits/854b37a05860fc95e195e664a6a23c4a8646090d to handle different bitstored values with a mask to get a  correct 16-bit representation in a short array.

The is a bug in my commit https://github.com/dcm4che/dcm4che/pull/616/commits/854b37a05860fc95e195e664a6a23c4a8646090d in line 721 (wrong bistored) => 
`int mask = (1 << imageDescriptor.getBitsStored());`

So with this fix and the commit https://github.com/dcm4che/dcm4che/pull/616/commits/3273f5338e881b15e0939bd93dce828db45190eb all the conversion must be correct with or without forcing 16-bit output. As described [here](https://github.com/dcm4che/dcm4che/issues/637#issuecomment-578536365), the output bitstored is of little importance except for JPEG encoders.

gunterze:
Could not verify:
> If you try to convert less than 16-bit signed pixels to a lossless syntax, the output data will be corrupted. Here are some samples:
http://www.dclunie.com/images/signedrange/index.html

E.g.:
```
$ dcm2dcm -C maxPixelValueError=0 --jpll ~/testdata/signedrangeimages /tmp | grep error:\ 0
Jan 27, 2020 4:27:00 PM org.opencv.osgi.OpenCVNativeLoader init
INFO: Successfully loaded OpenCV native library.
Jan 27, 2020 4:27:01 PM org.opencv.osgi.OpenCVNativeLoader init
INFO: Successfully loaded OpenCV native library.
16:27:01,020 DEBUG - Verified compressed frame #1 in 11 ms - max pixel value error: 0
16:27:01,047 DEBUG - Verified compressed frame #1 in 3 ms - max pixel value error: 0
..
16:27:03,273 DEBUG - Verified compressed frame #1 in 3 ms - max pixel value error: 0
```

Anyway, for JPEG 2000 lossy, I agree, that you have to perform sign extensions of the bits above BitsStored and the Bits per Sample applied by the encoder. =>  #648

nroduit:
Before your last commit https://github.com/dcm4che/dcm4che/commit/7ae55eeef8004ed8105409ccb6e3a44db77394b9 the images were corrupted with all lossless syntaxes (see image below).

The problem is that the verifier method only comparing the pixel value which is successful before and after the commit although the negative values have changed. The verification does not take into account whether the pixel values read by the decoder are correct.

Anyway, with https://github.com/dcm4che/dcm4che/commit/7ae55eeef8004ed8105409ccb6e3a44db77394b9 the issue can be closed.

Sélection_019

PR #617, opened 2020-01-13
[IEI-101092] Merging changes from QueryRetrieveLevel2 in master into …

homerocda:
…QueryRetrieveLevel to allow for lenient validation of CMove requests.

### DIFF
--- a/dcm4che-net/src/main/java/org/dcm4che3/net/service/QueryRetrieveLevel.java
+++ b/dcm4che-net/src/main/java/org/dcm4che3/net/service/QueryRetrieveLevel.java
@@ -38,12 +38,21 @@
 
 package org.dcm4che3.net.service;
 
+import org.dcm4che3.data.ElementDictionary;
 import org.dcm4che3.data.Tag;
 import org.dcm4che3.data.Attributes;
 import org.dcm4che3.data.IOD;
 import org.dcm4che3.data.VR;
 import org.dcm4che3.data.ValidationResult;
 import org.dcm4che3.net.Status;
+import org.dcm4che3.util.StringUtils;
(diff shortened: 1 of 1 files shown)


dcm4che-net/src/main/java/org/dcm4che3/net/service/QueryRetrieveLevel.java
```diff
@@ -96,6 +108,43 @@ public void validateRetrieveKeys(Attributes attrs,
         check(attrs.validate(retrieveKeysIOD(rootLevel, relational)));
     }
 
+    public static QueryRetrieveLevel validateRetrieveIdentifier(Attributes keys, Set<QueryRetrieveLevel> qrLevels,
+            boolean relational, boolean lenient) throws DicomServiceException {
+        return validateIdentifier(keys, qrLevels, relational, lenient);
+    }
+
+    private static QueryRetrieveLevel validateIdentifier(Attributes keys, Set<QueryRetrieveLevel> qrLevels,
+             boolean relational, boolean lenient) throws DicomServiceException {
```
craigtmoore: Why did you use a private method? It doesn't appear to be used anywhere else so it could be the body of the validateRetrieveIdentifier method

homerocda: It's copy/paste from https://github.com/dcm4che/dcm4che/blob/master/dcm4che-net/src/main/java/org/dcm4che3/net/service/QueryRetrieveLevel2.java.

Gunter used this structure to allow several overloads to use the same method. However, I agree this could be done in one public method.

Issue #449, opened 2019-05-01
Change DicomOutputStream.writeFileMetaInformation(fmi) to do not implicitly switch the Transfer Syntax


gunterze:
Change `DicomOutputStream.writeFileMetaInformation(fmi)` to do not implicitly switch the Transfer Syntax - particularly to do **not** wrap the `OutputStream` by `DeflaterOutputStream`.

patrick-peer:
This refactoring unfortunately breaks cases where the FMI is written separately from the dataset, and the transfer syntax is 1.2.840.10008.1.2 (Implicit VR) or 1.2.840.10008.1.2.2 (Big Endian). In these cases the corresponding booleans do not get switched anymore and, as a consequence, the DicomOutputStream encodes all followoing data elements wrongly.

I'd like to provide a fix by splitting up `DicomOutputStream.switchTransferSyntax(String)` to move the switching of `bigEndian` and `explicitVR` back to its old place and leave the setup of `DeflaterOutputStream` where it is now.
However, I am not quite sure about the implications of that, since I do not fully grasp the motivation for this issue in the first place. What do you think about the proposed fix?

patrick-peer:
Meanwhile I proposed pull request #619

PR #619, opened 2020-01-15
Split switchTransferSyntax() to make boolean toggle work in writeFileMetaInformation() again

patrick-peer:
The fix for issue #449 unfortunately breaks cases where the FMI is written separately from the dataset, and the transfer syntax is 1.2.840.10008.1.2 (Implicit VR) or 1.2.840.10008.1.2.2 (Big Endian). In these cases the corresponding booleans do not get switched any more and, as a consequence, the `DicomOutputStream` encodes all following data elements wrongly.

I am not 100% sure whether it is OK to split up `switchTransferSyntax()` for each and every use case, as I do not fully understand the motivation for issue #449. However, the proposed changes here do fix regressions on my end and maybe this is useful to you, too. In any case, please let me know if a more refined approach is required.


gunterze:
#642

PR #620, opened 2020-01-15
default return value for TransferSyntaxType.forUID() => null

patrick-peer:
I restored the old behaviour of `TransferSyntaxType.forUID()` to return `null` on unknown transfer syntax UIDs. At a certain point in my code I need the `TransferSyntaxType` to go on, but would rather stop processing with an error, instead of assuming a NATIVE format for unknown transfer syntaxes. The new implementation made this distinction impossible.

It seems to me, that this was quite likely a deliberate choice of yours, and that some of your code might depend on `TransferSyntaxType.forUID()` to never return `null`. I am proposing this pull request anyway, should this have been an oversight after all.


gunterze:
s. #641

PR #622, opened 2020-01-16
[IEI-101092] Aggregating errors before failing in QueryRetrieveLevel validation logic

homerocda:
The current QueryRetrieveLevel validation logic added in master fails at the first missing/invalid attribute found in the chain. The previous algorithm would aggregate all the wrong elements before throwing an exception with all the details. This change aggregates errors in the new validation logic and fixes the exception message string.


hczedik:
@homerocda Can you also open this as a PR for master and contribute is also there?

dcm4che-net/src/main/java/org/dcm4che3/net/service/UniqueKeyCheckFailureCollector.java
```diff
@@ -0,0 +1,97 @@
...
+    public void add(UniqueKeyCheckFailure failure) {
+        List<UniqueKeyCheckFailure> previous = failures.computeIfAbsent(failure.type, t -> new ArrayList<>());
+        previous.add(failure);
+    }
+
+    public boolean isEmpty() {
+        return failures.isEmpty();
+    }
+
+    public String getFailureMessage() {
```
hczedik: Minor suggestion: Unit tests might make sense for this method, as it is already quite complex.

dcm4che-net/src/main/java/org/dcm4che3/net/service/QueryRetrieveLevel.java
```diff
@@ -131,16 +132,23 @@ private static QueryRetrieveLevel validateIdentifier(Attributes keys, Set<QueryR
             throw invalidAttributeValue(Tag.QueryRetrieveLevel, value);
         }
 
+        UniqueKeyCheckFailureCollector collector = new UniqueKeyCheckFailureCollector();
```
hczedik: Also for the validateIdentifier(..) method unit tests would be nice.

Issue #387, opened 2018-12-12
Dcm2Jpg: create destination directory if directory doesn't exist


simur407:
**Description**
When first parameter is a directory of dcm files, and second parameter is name of destination directory, but destination directory does not exist, there is printed error message: "target <dest dir> is not a directory". I checked source code and found corresponding line:
`Dcm2Jpg.java:309 File dest = new File(argList.get(argc-1));`
`Dcm2Jpg.java:310 if ((argc > 2 || new File(argList.get(0)).isDirectory()) && !dest.isDirectory())`
There is check if `dest` is a directory, but `isDirectory()`, as we can see in [docs](https://docs.oracle.com/javase/7/docs/api/java/io/File.html#isDirectory()) will return `false` because 

> returns true if and only if the file denoted by this abstract pathname **exists** and is a directory; false otherwise

What's more, I found that it was indeed intended to allow non existing directories, because in lines:
`Dcm2Jpg.java:335 private void mconvert(File src, File dest) {`
`Dcm2Jpg.java:336       if (src.isDirectory()) {`
`Dcm2Jpg.java:337            dest.mkdir();`

There is call for `dest.mkdir()`

**To Reproduce**
1. Run dcm2jpg with first argument as real dicom directory path and second with some name the directory should have
For windows it will be `./dcm2jpg.bat <existing dir with dicom files> <some non existing dest dir>
2. See message "target <some non existing dest dir> is not a directory"

**Expected behavior**
Directory <some non existing dest dir> is created and all jpg files processed from source directory are saved there.

gunterze:
The current behavior, to require an existing target directory, but create sub-directory in it, is intended.

PR #649, opened 2020-01-28
Create directory if dest doesn't exist

ceachother:
This change fixes https://github.com/dcm4che/dcm4che/issues/387, create the directory to avoid exceptions thrown.


gunterze:
alternatively fixed by 6d434b4

Issue #647, opened 2020-01-27
NativeJ2kImageWriter: signed pixel values are compressed as unsigned


gunterze:
NativeJ2kImageWriter: signed pixel values are compressed as unsigned, which is a major issue for lossy compression. E.g.:

```
$ dcm2dcm -C maxPixelValueError=200 --j2ki ~/testdata/signedrangeimages/test16signed.dcm /tmp
Jan 27, 2020 10:24:57 PM org.opencv.osgi.OpenCVNativeLoader init
INFO: Successfully loaded OpenCV native library.
22:24:57,706 DEBUG - Compressor: org.dcm4che3.opencv.NativeJ2kImageWriter
22:24:57,709 DEBUG - Verifier: org.dcm4che3.opencv.NativeImageReader
22:24:57,821 DEBUG - Compressed frame #1 in 82 ms, ratio 7.3845463:1
Jan 27, 2020 10:24:57 PM org.opencv.osgi.OpenCVNativeLoader init
INFO: Successfully loaded OpenCV native library.
WARN: Force to use signed data as defined in DICOM
22:24:57,850 DEBUG - Verified compressed frame #1 in 28 ms - max pixel value error: 65535
Failed to transcode /home/gunter/testdata/signedrangeimages/test16signed.dcm: Decompressed pixel data differs up to 65535 from original pixel data
org.dcm4che3.imageio.codec.CompressionVerificationException: Decompressed pixel data differs up to 65535 from original pixel data
	at org.dcm4che3.imageio.codec.Transcoder.verify(Transcoder.java:864)
	at org.dcm4che3.imageio.codec.Transcoder.compressFrame(Transcoder.java:648)
	at org.dcm4che3.imageio.codec.Transcoder.compressPixelData(Transcoder.java:492)
	at org.dcm4che3.imageio.codec.Transcoder.processPixelData(Transcoder.java:429)
	at org.dcm4che3.imageio.codec.Transcoder.access$300(Transcoder.java:70)
	at org.dcm4che3.imageio.codec.Transcoder$1.readValue(Transcoder.java:387)
	at org.dcm4che3.io.DicomInputStream.readAttributes(DicomInputStream.java:534)
	at org.dcm4che3.imageio.codec.Transcoder.transcode(Transcoder.java:366)
	at org.dcm4che3.tool.dcm2dcm.Dcm2Dcm.transcodeWithTranscoder(Dcm2Dcm.java:411)
[...]

Issue #637, opened 2020-01-23
NativeJLSImageWriter and NativeJ2kImageWriter: consider passed value of Bits per Sample


gunterze:
`org.dcm4che3.opencv.NativeJLSImageWriter` and `org.dcm4che3.opencv.NativeJ2kImageWriter` ignore passed value of Bits per Sample by 
```java
params[Imgcodecs.DICOM_PARAM_BITS_PER_SAMPLE] = desc.getBitsCompressed(); // Bits per sample
```
but always include all 16 bits into the compressed stream for images with Bits Allocated = 16.

From added TranscoderTests:
```java
    @Test
    public void testCompress12BitsJPLL() throws Exception {
        test("MR2_UNC", "MR2_UNC-JPLL.dcm", UID.JPEGLossless, true);
        assertEquals(12, jpegBitsPerSample("MR2_UNC-JPLL.dcm"));
    }

    @Test
    public void testCompress12BitsJLSL() throws Exception {
        test("MR2_UNC", "MR2_UNC-JLSL.dcm", UID.JPEGLSLossless, true);
        assertEquals(12, jpegBitsPerSample("MR2_UNC-JLSL.dcm"));
    }

    @Test
    public void testCompress12BitsJ2KR() throws Exception {
        test("MR2_UNC", "MR2_UNC-J2KR.dcm", UID.JPEG2000LosslessOnly, true);
        assertEquals(12, jpegBitsPerSample("MR2_UNC-J2KR.dcm"));
    }

```
last 2 fail with
```
java.lang.AssertionError: 
Expected :12
Actual   :16
```

nroduit:
The JPEG-LS encoder handles 8/16-bit data.
The JPEG2000 encoder handles 8/16/32-bit data
So it should be OK to get 16-bit from the header.

The JPG encoders (with a specific 12-bit encoder):
1) 1.2.840.10008.1.2.4.50: Lossy compression with JPEG baseline => 8 bit
2) 1.2.840.10008.1.2.4.51: Lossy compression with JPEG extended sequential => 8/12 bit
3) 1.2.840.10008.1.2.4.53: Lossy compression with JPEG spectral selection => 8/12 bit
4) 1.2.840.10008.1.2.4.55: Lossy compression with JPEG full progression => 8/12 bit
5) 1.2.840.10008.1.2.4.57: Pseudo-Lossless JPEG compression => 8/12/16 bit
[...]

Issue #638, opened 2020-01-23
Include unused bits of signed pixel values in compressed stream by default


gunterze:
Use negative values for _Image Write Param_ `bitsCompressed` to indicate, to apply its absolute value only on signed pixel data - and only include number of bits specified by *Bits Stored* for unsigned pixel data.

Set `bitsCompressed=-16` for JPEG Lossless, JPEG LS and JPEG 2000 Compression in `ImageWriterFactory.properties`
```
1.2.840.10008.1.2.4.57:jpeg-cv:org.dcm4che3.opencv.NativeJPEGImageWriter::bitsCompressed=-16;compressionType=LOSSLESS-6
1.2.840.10008.1.2.4.70:jpeg-cv:org.dcm4che3.opencv.NativeJPEGImageWriter::bitsCompressed=-16;compressionType=LOSSLESS-1
1.2.840.10008.1.2.4.80:jpeg-ls-cv:org.dcm4che3.opencv.NativeJLSImageWriter::bitsCompressed=-16
1.2.840.10008.1.2.4.81:jpeg-ls-cv:org.dcm4che3.opencv.NativeJLSImageWriter::bitsCompressed=-16;nearLossless=2
1.2.840.10008.1.2.4.90:jpeg2000-cv:org.dcm4che3.opencv.NativeJ2kImageWriter::bitsCompressed=-16;compressionType=LOSSLESS
1.2.840.10008.1.2.4.91:jpeg2000-cv:org.dcm4che3.opencv.NativeJ2kImageWriter::bitsCompressed=-16
```

Can be disabled by
```
Property[] compressParams = new Property("bitsCompressed",0);
Transcoder.setCompressParams(compressParams)
```

Extends #599

PR #659, opened 2020-02-03
Update to native lib 4.2.0-dcm

nroduit:
* Update to OpenCV 4.2.0
* Update to charLS 2.1.0 (API has changed)
* Minor update of OpenJPEG and JPEG
* Fix sign data for jpeg-2000 lossy compression #647
* Fix consider bit per sample for compression of JPEG-LS and JPEG-2000 #637
* Add image content hash (will be used for tests see this [page]( http://qtandopencv.blogspot.com/2016/06/introduction-to-image-hash-module-of.html)) 
* Replace lossy quality parameter form psnr to ratio factor (which is easier to understand)

Here is a summary of the benefits of using bit precision for compression:

**JPEG-LS**: desc.getBitsCompressed() => if betwween 2 and 16 included the value is applied otherwise is either 8 or 16 according to the input image.
The percentage gain of 9-15 vs 16 bitcompressed (lossless and lossy, only for unsigned data): 10 to 25%
signed data: force to 16 bitcompressed without any gain

**JPEG-2000**: also 2 to 16 : 
For lossless, the gain is not significant (~1%)
The percentage gain of 12 vs 16 bitcompressed (lossy for signed and unsigned data): 20 to 25% 
The percentage gain of 9 vs 16 bitcompressed (lossy for signed and unsigned data): 40 to 50%

**JPEG-LOSSLESS**:
the gain is not significant (~1%)


nroduit:
I have tested all decoders and encoders on a large and varied data set.

dcm4che-imageio-opencv/src/main/java/org/dcm4che3/opencv/NativeJLSImageWriter.java
```diff
@@ -115,14 +115,15 @@ public void write(IIOMetadata streamMetadata, IIOImage image, ImageWriteParam pa
...
                 int channels = CvType.channels(cvType);
-                int dcmFlags = CvType.depth(cvType) == CvType.CV_16S ? Imgcodecs.DICOM_IMREAD_SIGNED
-                    : Imgcodecs.DICOM_IMREAD_UNSIGNED;
-                if((dcmFlags & Imgcodecs.DICOM_IMREAD_SIGNED) == Imgcodecs.DICOM_IMREAD_SIGNED) {
+                boolean signed = desc.isSigned();
+                int dcmFlags = signed ? Imgcodecs.DICOM_FLAG_SIGNED : Imgcodecs.DICOM_FLAG_UNSIGNED;
+                if(signed) {
                     LOGGER.warn("Force compression to JPEG-LS lossless as lossy is not adapted to signed data.");
                     jpeglsNLE = 0;
+                    bitCompressed = 16; // Extend to bit allocated to avoid exception as negative values are treated as large positive values
```
gunterze: Is already handled by https://github.com/dcm4che/dcm4che/issues/638 - which allows to change the default behavior and restrict the compression to Bit Stored also for signed data, relying that receiving applications behaves according [DICOM PS 3.5, Pixel Data Encoding of Related Data Elements](dicom.nema.org/medical/dicom/current/output/html/part05.html#sect_8.1.1)
> Receiving applications may not assume anything about the contents of unused bits, and in particular may not assume that they are zero, or that they contain sign extension bits.

nroduit:
Yes, I know but I wanted to force in any case because an error is thrown by charls.

PR #664, opened 2020-02-07
[IEI-103029] Fixing thread leak when writing a dimse response fails

homerocda:
When a device does not respect the DICOM standard and sends messages asynchronously even though we told them to behave synchronously, there might be a thread leak when writing the response fails and there are other threads waiting for a spot to add a message to the response queue.

Removing the message from the queue when an exception is thrown fixes the issue.


hczedik:
Get this reviewed in parallel by Gunter on the master branch please too.

PR #673, opened 2020-02-12
Merge IEI-93836 - Properly escape special characters in the Issuer and IDWithIssuer classes

qzhang0118:


hczedik:
@gunterze This is a change we discussed long time ago via email, to support issuers with special characters better.

gunterze:
Solved by #674

PR #682, opened 2020-02-21
[IEI-104094] Fix NPE when store fails + add error details

homerocda:


hczedik:
Just a note: Would be good to pull this class in-house (into one of our server-side components), as it is not on the master branch of dcm4che.

homerocda:
> 
> 
> Just a note: Would be good to pull this class in-house (into one of our server-side components), as it is not on the master branch of dcm4che.

Thanks for the observation. We have created a Task to check the impact of pulling this in.

PR #689, opened 2020-03-05
Update DicomInputStream#readValue()

craigtmoore:
remove the code that creates multiple arrays in memory until everything fits and instead rely on the length value. We're proposing this because we are seeing OOM exceptions due to the way this code handles tags with excessively large values in them.

### DIFF
--- a/dcm4che-core/src/main/java/org/dcm4che3/io/DicomInputStream.java
+++ b/dcm4che-core/src/main/java/org/dcm4che3/io/DicomInputStream.java
@@ -766,7 +766,7 @@ public class DicomInputStream extends FilterInputStream
 
         } catch (IOException e) {
 
-            if (Objects.isNull(e.getMessage()) && Objects.nonNull(e.getCause())) {
+            if ((Objects.isNull(e.getMessage()) || e.getMessage().equals("")) && Objects.nonNull(e.getCause())) {
 
                     throw new IOException(String.format("IOException during read of %s #%d @ %d",
                             TagUtils.toString(tag), length, tagPos), e.getCause());
--- a/dcm4che-core/src/main/java/org/dcm4che3/io/DicomInputStream.java
+++ b/dcm4che-core/src/main/java/org/dcm4che3/io/DicomInputStream.java
@@ -752,7 +752,7 @@ public class DicomInputStream extends FilterInputStream
                 throw new EOFException("DicomInputStream length is less than zero");
             }
 
-            if (length > allocateLimit) {
+            if (allocateLimit > 0 && length > allocateLimit) {
                 LOG.warn("Expected value length {} exceeds allocate limit {}", length, allocateLimit);
             }
 
--- a/dcm4che-core/src/main/java/org/dcm4che3/io/DicomInputStream.java
+++ b/dcm4che-core/src/main/java/org/dcm4che3/io/DicomInputStream.java
@@ -162,8 +162,6 @@ public class DicomInputStream extends FilterInputStream
     }
 
     /** 
-     * Returns the limit of initial allocated memory for element values.
-     * 
      * By default, the limit is set to 67108864 (64 MiB).
      *
      * @return Limit of initial allocated memory for value or -1 for no limit
@@ -174,22 +172,9 @@ public class DicomInputStream extends FilterInputStream
     }
 
     /**
-     * Sets the limit of initial allocated memory for element values. If the
-     * value length exceeds the limit, a byte array with the specified size is
-     * allocated. If the array can filled with bytes read from this
-     * <code>DicomInputStream</code>, the byte array is reallocated with
-     * twice the previous length and filled again. That continues until
-     * the twice of the previous length exceeds the actual value length. Then
-     * the byte array is reallocated with actual value length and filled with
-     * the remaining bytes for the value from this <code>DicomInputStream</code>.
-     * 
-     * The rational of the incrementing allocation of byte arrays is to avoid
-     * OutOfMemoryErrors on parsing corrupted DICOM streams.
-     * 
      * By default, the limit is set to 67108864 (64 MiB).
      * 
      * @param allocateLimit limit of initial allocated memory or -1 for no limit
-     * 
      */
     public final void setAllocateLimit(int allocateLimit) {
         this.allocateLimit = allocateLimit;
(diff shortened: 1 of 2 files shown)


dcm4che-core/src/main/java/org/dcm4che3/io/DicomInputStream.java
```diff
@@ -760,27 +745,30 @@ private void readFragments(Attributes attrs, int fragsTag, VR vr)
...
+
             return value;
+
         } catch (IOException e) {
-            LOG.warn("IOException during read of {} #{} @ {}",
-                    TagUtils.toString(tag), length, tagPos, e);
-            throw e;
+
+            throw new IOException(String.format("IOException during read of %s #%d @ %d",
+                    TagUtils.toString(tag), length, tagPos), e);
```
mdoris: I would prefer the original exception (or its cause) to be thrown instead of wrapping the exception with another IOException.

dcm4che-core/src/main/java/org/dcm4che3/io/DicomInputStream.java
```diff
@@ -760,27 +745,30 @@ private void readFragments(Attributes attrs, int fragsTag, VR vr)
...
-                int newLength = Math.min(valLen, allocLen << 1);
-                value = Arrays.copyOf(value, newLength);
-                readFully(value, allocLen, newLength - allocLen);
-                allocLen = newLength;
+
+            if (length < 0) {
+                throw new EOFException("DicomInputStream length is less than zero");
+            }
+
+            if (length > allocateLimit) {
```
homerocda: Looks like the previous code allowed for allocateLimit = 0 to be interpreted as "no limit". If this continues as the expected behavior, then we might see unnecessary warnings for every alocation.

dcm4che-core/src/main/java/org/dcm4che3/io/DicomInputStream.java
craigtmoore: Ok, I've updated it.

dcm4che-core/src/main/java/org/dcm4che3/io/DicomInputStream.java
homerocda: Correction: unlimited allocation was "-1", not 0.

craigtmoore: OK, I'll update it to first check that the allocationLimit is non-zero.

gunterze:
The change breaks the graceful handling of faulty DICOM streams, detectable by a decoded value length larger than the remaining bytes in the stream

craigtmoore:
That may be, but it also introduces a memory leak as it churns through
copies of the array to find the "best fit". We've had OOM exceptions for
this code at sites. The same code exists in version 2 and we have reports
about it there as well. If you have a better approach, then please let me
know.

gunterze:
```java
                value = Arrays.copyOf(value, newLength);
```
does **not** cause a memory leak, because no reference to the previous `byte[]` is kept!

mdoris:
Not having a reference is not the same as the memory being freed. It remains in memory until the GC gets to it.

We have had multiple OOM where the thread that triggers the OOM is performing the Arrays.copy() call. The heap dumps generated by the OOM are showing 70% of the system memory consumed with byte[]s and further examination is showing 5-6 copies of identically sized, identically values byte[]s.

The original code does not provide any protection against OOM as the return value of the original code is ALWAYS a byte[] or equal (or greater in older versions) size to the length. The new code does the same thing except in one read. In both cases if there is an attempt to read past the end of the stream an IOException is thrown.

gunterze:
The old code prevents the OOM in case of parsing an invalid 4 byte length value, the new code tries to allocate a byte[] anyway - so the OOM happens before detecting EOF!

mdoris:
How is that any different from the old code? 

You will still end up creating a byte[] of the size of the valLen either through direct allocation or via the Arrays.copyOf(). None of this actually prevents this byte[] from being allocated. The only difference is the old code creates multiple unreferenced byte[]s of varying sizes until it finally allocates the byte[valLen] (or even bigger depending on the bitshifting).

gunterze:
In the old code, a new byte[] with doubled length is only allocated, if the previous byte[] could be filled by reading from the stream...

craigtmoore:
What we should have instead is a way to detect a faulting length value or a solution that can detect the actual length of the value in the presence of a faulty length value.  The existing solution doesn't deal with that well.

mdoris:
Yes, I can read that it does that, but how is that effectively better?

Assuming the allocLen = 1/2 the valLen:

In the old version, if there is no data at valLen - x, but there is at valLen/2, you get an IOException reading the stream the 2nd read. In the current version, you get an IOException reading the stream on the first read.

In the old version, if there is no data at valLen-2 and data at valLen/2, the old version has consumed 1.5*valLen bytes of memory vs the current version consuming valLen.

In the old version, if there was no data (valLen/2) - 2 the old version throws an IOException and consumes valLen/2 bytes memory. The new version throws an IOException and consumed valLen bytes memory.

We know this code OOMs performing the Arrays.copyOf. We have the heapdumps/thread dumps to prove this. We are intentionally exchanging some error handling of minor value (as it doesn't prevent or correct the error, it simply provides arguably more efficient memory use in failure cases) for a simpler approach that does not consume nearly as much memory in the functional path.

hczedik:
Mike, Craig:
The problem is that for faulty data the length value could be really huge, immediately causing an OOM. The old code was preventing that, in the case the actual data from the stream is shorter.
But I tend to agree that is not optimal to make the functional normal case less efficient, to handle faulty data.

Anyways, if you just want to disable the allocateLimit, why not just set DEF_ALLOCATE_LIMIT to -1? Why this whole change?

mdoris:
Explain how a faulty length value does not OOM in the original implementation please.

hczedik:
If the actual data that can be read (with readFully()) is shorter, than it will at most read the whole file, but then stop with an IOException.

mdoris:
except that in that scenario, this original code has already OOM'd due to the copies of the byte[].

If we change this implementation to not immediately fully allocate, but eliminate the Arrays.copyOf and the bitshifting, would this then be acceptable?

hczedik:
No, the old code would not have necessarily OOM'd.
Imagine you are reading a 1mb file. It contains a faulty length of 2GB. The old code would only allocate 64mb and nothing more. With your change you will allocate the full 2GB and are more likely to run out of memory.
(But I have to admit it is a very specific case this is solving. Obviously you can easily construct other examples, where the old code will allocate more memory.)

How do you want to eliminate the Arrays.copyOf() without fully allocating?

mdoris:
I am tinkering with the idea of a 

int listSize = Math.ceil(length/allocLen);
List byteArrays = LinkedList<byte[]>(listSize);
for (0 to listSize) {
   byte[] readBytes = read(allocLength || remaining); // remaining calculated for last iteration
   byteArrays.add(readBytes);
}

At the end I don't see a way to avoid a temporary second copy of the data in order to build the array to be returned... but as I mentioned, this is a very early concept.

gunterze:
https://github.com/dcm4che/dcm4che/issues/691

mdoris:
The error that we are dealing with actually involves valid lengths and valid data. In those scenarios in was hitting OOM trying to resize the array.

However, it is important to not OOM on the invalid case as well.

hczedik:
Just because the OOM was raised for this method, doesn't mean it is the root cause of the memory shortage. It could also be that your heap is filled up with other stuff, and just by coincidence this allocation triggered the OOM.
What is the use case where you need to load these tags into memory? How large is the length of the tag that it is choking on? How much memory did the server have in total configured?
(Okay we should be moving this discussion to our internal tracker :-) )

PR #692, opened 2020-03-11
ReadableAttributes interface that can be used as read-only view of Attributes

hczedik:
Hi @gunterze,
this is an idea that we discussed some time ago.
The goal is to have a read-only view of Attributes, which can be used to document that a certain dataset should not be modified. That is very useful in a multi-threaded application, and also if different external teams are adding plug-ins to an existing platform.

The implementation is not finished. I still want to change also the getSequence() methods in the interface to return a ReadableSequence, which cannot be (easily) modified.

But before I put further work into this, I wanted to check with you, if you would accept these changes, once they are finished?


hczedik:
Alternatively implemented by https://github.com/dcm4che/dcm4che/commit/a3cf20f275717e13ce64e0cc92b6f0c9cec9a6e7

PR #701, opened 2020-03-30
Actually indent the output with whitespace if option -I,--indent is given for dcm2xml tool

hczedik:


hczedik:
@gunterze Did you have a look at this already? It is a tiny change, but it leads to much nicer output of the tool.

gunterze:
It's on my list - if not today, than tomorrow.

Sent with [ProtonMail](https://protonmail.com) Secure Email.

‐‐‐‐‐‐‐ Original Message ‐‐‐‐‐‐‐

hczedik:
True, such a CLI parameter might also be useful. But we can always still add it later.
I've merged the branch.

PR #749, opened 2020-07-09
Update Attributes.java

satheesh09:
Increasing length of STRING_WIDTH as the DICOM tag names at the end of the command line output are getting stripped, when the StudyInstanceUID is longer.


gunterze:
Rejected. The default 78 is motivated by the still common default terminal width of 80. https://stackoverflow.com/questions/4651012/why-is-the-default-terminal-width-80-characters

PR #750, opened 2020-07-14
Internal config change event add changed paths

stephenfrederick:


dcm4che-conf/dcm4che-conf-core-api/src/main/java/org/dcm4che3/conf/core/api/InternalConfigChangeEvent.java
```diff
@@ -51,5 +53,20 @@
  * @author Alexander Hoermandinger <alexander.hoermandinger@agfa.com>
  */
 public class InternalConfigChangeEvent implements Serializable {
+    private static final long serialVersionUID = -7690181293743297960L;
+
+    private final List<String> changedPaths;
+  
+    public InternalConfigChangeEvent() {
+        changedPaths = new ArrayList<>();
```
hczedik, line 61: Maybe Collections.emptyList() instead? (Not important.)

dcm4che-conf/dcm4che-conf-core-api/src/main/java/org/dcm4che3/conf/core/api/InternalConfigChangeEvent.java
```diff
@@ -51,5 +53,20 @@
...
+    }
+    
+    public InternalConfigChangeEvent(List<String> changedPaths) {
+        this.changedPaths = changedPaths;
```
hczedik, line 65: Maybe you want to make this unmodifiable and check for null? (something like `this.changedPaths = Collections.unmodifiableList(Objects.requireNonNull(changedPaths))`). (Not important.)

PR #773, opened 2020-08-12
Introduce a static thread pool in AuditLogger to fix thread leak

qzhang0118:


dcm4che-net-audit/src/main/java/org/dcm4che3/net/audit/NamedThreadFactory.java
```diff
@@ -0,0 +1,72 @@
...
+    private AtomicInteger threadCounter;
+    private final String namePrefix;
+
+    /**
+     * Create a new NamedThreadFactory that uses the indicated namePrefix to
+     * generate thread names.
+     * @param namePrefix Prefix to use when generating thread names. NULL/Empty will use a default value.
+     */
+    public NamedThreadFactory(String namePrefix) {
+        if (namePrefix == null || namePrefix.length() == 0) {
```
hczedik, line 56: Suggestion (not important): `namePrefix.isEmpty()` instead of `namePrefix.length() == 0`.

dcm4che-net-audit/src/main/java/org/dcm4che3/net/audit/NamedThreadFactory.java
```diff
@@ -0,0 +1,72 @@
...
+ *
+ * ***** END LICENSE BLOCK ***** */
+
+package org.dcm4che3.net.audit;
+
+import java.util.concurrent.ThreadFactory;
+import java.util.concurrent.atomic.AtomicInteger;
+
+public class NamedThreadFactory implements ThreadFactory {
+    public static final String DEFAULT_PREFIX = "audit-default- ";
```
hczedik, line 45: Is the space character at the end of this string on purpose?

dcm4che-net-audit/src/main/java/org/dcm4che3/net/audit/AuditLogger.java
```diff
@@ -1518,7 +1521,7 @@ private void startIdleTimer() {
                 try {
                     Device device = conn.getDevice();
                     if (device.getScheduledExecutor() == null) {
-                        device.setScheduledExecutor(Executors.newSingleThreadScheduledExecutor());
+                        device.setScheduledExecutor(scheduledExecutor);
                     }
                     idleTimer = device.schedule(
                             new Runnable() {
```
KevinDougan, line 1527: Should we also add a call to scheduledExecutor.shutdown() inside the onIdleTimerExpired() method below to clean up the Executor?

greg-knox: No - then the thread pool is shutdown and can't be re-used by the next call which would use the idle timer

dcm4che-net-audit/src/main/java/org/dcm4che3/net/audit/NamedThreadFactory.java
qzhang0118: it has been changed to namePrefix.isEmpty()

dcm4che-net-audit/src/main/java/org/dcm4che3/net/audit/NamedThreadFactory.java
qzhang0118: the space has been removed

dcm4che-net-audit/src/main/java/org/dcm4che3/net/audit/AuditLogger.java
```diff
@@ -94,6 +95,8 @@
 
     private static Disruptor<AuditMessageEvent> disruptor;
 
+    private static final ScheduledExecutorService scheduledExecutor = Executors.newSingleThreadScheduledExecutor(new NamedThreadFactory(null));
```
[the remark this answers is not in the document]
qzhang0118: This single thread was first added to fix many warnings "Failed to start Idle timeout" from AuditLogger due to "java.lang.IllegalStateException: scheduled executor service not initalized"
https://github.com/dcm4che/dcm4che/pull/680/commits/4a69ac7e3358744a0bfae96c1941559100068478

PR #790, opened 2020-09-17
Change log levels to trace to avoid verbose logging

alan-farquhar:


dcm4che-core/src/main/java/org/dcm4che3/data/Attributes.java
```diff
@@ -753,7 +753,7 @@ public Sequence getSequence(String privateCreator, int tag) {
             
             return vr.toBytes(value, getSpecificCharacterSet(vr));
         } catch (UnsupportedOperationException e) {
-            LOG.info("Attempt to access {} {} as bytes", TagUtils.toString(tag), vr);
+            LOG.trace("Attempt to access " + TagUtils.toString(tag) + " " + vr + " as bytes",e);
```
craigtmoore, line 756: You can leave the 'e' and use the {} in log4j.

gunterze:
An attempt to get the value of a Sequence attribute as byte[], String, int, float or double is a coding error in the client and should be logged on INFO level.

dcm4che-core/src/main/java/org/dcm4che3/data/Attributes.java
mdoris: I don't think the stack from the exception gets logged correctly using the {}.

mdoris (approved):

gunterze:
Will not agree to merge it into master!

mdoris:
> An attempt to get the value of a Sequence attribute as byte[], String, int, float or double is a coding error in the client and should be logged on INFO level.

The problem is not that this is a bad call to toBytes[] etc. the problem is that this code is setup to use exception handling as flow control through the Attributes.getBytes() where it performs an instanceof Value check knowing that multiple implementations of Value intentionally throw UnsupportedOperationException. This exception is being thrown multiple times during the reading of Pixel data which is expected to be able to be obtained as bytes.

Additionally, when this exception is thrown, not only is it proven to be very noisy in many site locations the log message itself provides no actual useful information. It simply says "xxx tag accessed as bytes", it presents no information as to why this fails.  If it actually provided diagnostic information about an error it would be useful at the debug level, as it this log statement is only useful to determine that this code was touched and provides information that is only useful when tracing the execution path.

gunterze:
It's stack-trace points you to the erroneous client code.

mdoris:
> It's stack-trace points you to the erroneous client code.

Stack trace? Our change adds a stack trace. The original code supressed it.

gunterze:
You may add the stacktrace - but please complaint with slf4j conventions - but I reject the change of the logging level.

dcm4che-core/src/main/java/org/dcm4che3/data/Attributes.java
hczedik: Craig is right. If you do it like this:
`LOG.info("Attempt to access {} {} as bytes", TagUtils.toString(tag), vr, e);`
The stack trace should get correctly logged.

hczedik:
Maybe you want to keep the info level logging and additionally add trace level logging which also prints the stack trace?

alan-farquhar:
@hczedik  the info logging is already flooding our logs at sites as it is without the stack trace.

hczedik (changes requested):
Use the slf4j logging format with "{}".

alan-farquhar:
Not going forward with this pull request

PR #792, opened 2020-09-17
Reduce the logging for unsupportedOperationException to trace to avoi…

alan-farquhar:
The logging in the attributes class in the unsupportedOperationException is causing the server logs to fill up.

### DIFF
--- a/dcm4che-core/src/main/java/org/dcm4che3/data/Attributes.java
+++ b/dcm4che-core/src/main/java/org/dcm4che3/data/Attributes.java
@@ -678,7 +678,7 @@ public class Attributes implements Serializable {
             
             return vr.toBytes(value, getSpecificCharacterSet(vr));
         } catch (UnsupportedOperationException e) {
-            LOG.info("Attempt to access {} {} as bytes", TagUtils.toString(tag), vr);
+            LOG.trace("Attempt to access {} {} as bytes", TagUtils.toString(tag), vr, e);
             return null;
         }
     }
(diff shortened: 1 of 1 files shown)


craigtmoore (approved):
Looks good to me

mdoris (approved):

Bronkoknorb:
It's a problematic change, because Gunter won't accept it on master, bringing us further away from it.
I'd like to understand why it is necessary. Are we ingesting so much corrupt data? Or are we misusing the library (calling the wrong method)?
I haven't seen the same logging a lot in the Desktop client log (but maybe I have just not yet payed attention to it).
(Please also point me to the Jira issue in Teams.)
(By the way, why the lowest "trace" level and not "debug"?)

alan-farquhar:
hi Hermann,

It appears as though the dicom that we're ingesting is leading to this type of message appearing quite regularly in our logs.  For example if the pixel data tag is normal then we don't get the log message:

image

However if the pixel data tag is a sequence then we run into this log message: 

image

The defect is here: 

http://jiraprod.agfahealthcare.com/browse/IEI-119520

gunterze:
```java
Objects value = attrs.getValue(Tag.PixelData);
if (value instanceof byte[]) {
  byte[] byteValue = (byte[]) value;
...
```
Generally, you should avoid to load pixel data into memory, better to stream it directly from network or file...

It would have been better to just don't catch the UnsupportedOperationException, so clients would get aware of the problem immediately. Hiding the log message further reduce the visibility of the problem, without solving it.

hczedik:
Thanks, @gunterze!

@alan-farquhar, to extend a bit on what Gunter wrote: The pixel data in your second example is not a sequence, but is encapsulated pixel data, which is always the case if the data is encoded (compressed). It cannot be read into a single byte array.
See also http://dicom.nema.org/dicom/2013/output/chtml/part05/sect_A.4.html
and https://github.com/dcm4che/dcm4che/blob/master/dcm4che-core/src/main/java/org/dcm4che3/data/Fragments.java#L52
(Gunter is right that we should always avoid reading the pixel data into memory, but stream it whenever possible.)

So I think we can close the PR, keep the logging as it is, and better go for improving our usage of the library.

alan-farquhar:
Yes thanks @gunterze and @hczedik we did locate the client code which is using the library incorrectly.  So i will close this pull request without merging.

PR #814, opened 2020-11-06
IEI-122301 Create DicomConfiguration#listDevices for faster bulk searches

RhuanKrum:
Create DicomConfiguration#listDevices for faster bulk searches

- Introduce DicomConfiguration#listDevices for bulk device searches
- Introduce Configuration#getConfigurationNodes for bulk node searches

I am attaching below a few examples of performance improvement using 10, 100 and 1000 devices. The performance will progressively increase given the amount of devices to be listed:

#### 10 devices (listDevices vs findDevice)
IEI-122301 - dcm4che 10

#### 100 devices (listDevices vs findDevice)
IEI-122301 - dcm4che 100

#### 1000 devices (listDevices vs findDevice)
IEI-122301 - dcm4che 1000

### DIFF
--- a/dcm4che-conf/dcm4che-conf-core/src/main/java/org/dcm4che3/conf/core/storage/InMemoryReadOnlyConfiguration.java
+++ b/dcm4che-conf/dcm4che-conf-core/src/main/java/org/dcm4che3/conf/core/storage/InMemoryReadOnlyConfiguration.java
@@ -45,8 +45,8 @@ import org.dcm4che3.conf.core.api.ConfigurationException;
 import org.dcm4che3.conf.core.Nodes;
 import org.dcm4che3.conf.core.api.Path;
 
-import java.util.Iterator;
-import java.util.Map;
+import java.util.*;
+import java.util.stream.Collectors;
 
 /**
  * @author Roman K
@@ -67,6 +67,13 @@ public class InMemoryReadOnlyConfiguration implements Configuration {
     @Override
     public Object getConfigurationNode(Path path, Class configurableClass) throws ConfigurationException {
         return Nodes.getNode(root, path.getPathItems());
+    }
+
+    @Override
+    public List<Object> getConfigurationNodes(Class configurableClass, Path... paths) throws ConfigurationException {
+        return Arrays.stream(paths)
+                .map(path -> getConfigurationNode(path, configurableClass))
+                .collect(Collectors.toList());
     }
 
     @Override
--- a/dcm4che-conf/dcm4che-conf-core/src/main/java/org/dcm4che3/conf/core/storage/InMemoryReadOnlyConfiguration.java
+++ b/dcm4che-conf/dcm4che-conf-core/src/main/java/org/dcm4che3/conf/core/storage/InMemoryReadOnlyConfiguration.java
@@ -42,47 +42,44 @@ package org.dcm4che3.conf.core.storage;
 
 import org.dcm4che3.conf.core.api.Configuration;
 import org.dcm4che3.conf.core.api.ConfigurationException;
-import org.dcm4che3.conf.core.Nodes;
 import org.dcm4che3.conf.core.api.Path;
 
 import java.util.*;
-import java.util.stream.Collectors;
 
 /**
  * @author Roman K
  */
 public class InMemoryReadOnlyConfiguration implements Configuration {
 
-    private final Map<String, Object> root;
+    private final InMemoryConfiguration delegate;
 
     public InMemoryReadOnlyConfiguration(Map<String, Object> root) {
-        this.root = root;
+        this.delegate = new InMemoryConfiguration(root);
     }
 
     @Override
     public Map<String, Object> getConfigurationRoot() throws ConfigurationException {
-        return root;
+        return delegate.getConfigurationRoot();
     }
 
     @Override
     public Object getConfigurationNode(Path path, Class configurableClass) throws ConfigurationException {
-        return Nodes.getNode(root, path.getPathItems());
+        return delegate.getConfigurationNode(path, configurableClass);
     }
 
     @Override
     public List<Object> getConfigurationNodes(Class configurableClass, Path... paths) throws ConfigurationException {
-        return Arrays.stream(paths)
-                .map(path -> getConfigurationNode(path, configurableClass))
-                .collect(Collectors.toList());
+        return delegate.getConfigurationNodes(configurableClass, paths);
     }
 
     @Override
     public boolean nodeExists(Path path) throws ConfigurationException {
-        return Nodes.nodeExists(root, path.getPathItems());
+        return delegate.nodeExists(path);
     }
 
     @Override
     public void refreshNode(Path path) throws ConfigurationException {
+        delegate.refreshNode(path);
     }
 
     @Override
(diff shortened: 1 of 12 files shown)


dcm4che-conf/dcm4che-conf-dicom/src/test/java/org/dcm4che3/conf/dicom/DicomConfigurationTest.java
```diff
@@ -91,6 +93,88 @@ public void renameAETest() throws ConfigurationException {
...
+
+        Device originalDevice = createDeviceWithArchiveExtension("DEVICE", "Default", "Other", "Storage", DeviceType.ARCHIVE, DeviceType.PRINT);
+        config.persist(originalDevice);
+
+        Device[] devices = config.listDevices(DeviceType.ARCHIVE);
+        Assert.assertTrue(devices.length == 1);
+
+        Device device = devices[0];
+
+        Assert.assertTrue(deepCompare(originalDevice, device));
```
craigtmoore, line 109: I would prefer you add a message to each `Assert.assertTrue` statement about what failed. Otherwise, if there is a test failure, then the message returned will be, "True is False". So for each of these statements you should include something like, "Original device does not match the device returned by listDevices".

dcm4che-conf/dcm4che-conf-dicom/src/test/java/org/dcm4che3/conf/dicom/DicomConfigurationTest.java
```diff
@@ -91,6 +93,88 @@ public void renameAETest() throws ConfigurationException {
...
+        Device originalDevice = createDeviceWithArchiveExtension("DEVICE", "Default", "Other", "Storage", DeviceType.ARCHIVE, DeviceType.PRINT);
+        config.persist(originalDevice);
+
+        Device[] devices = config.listDevices(DeviceType.ARCHIVE);
+        Assert.assertTrue(devices.length == 1);
+
+        Device listedDevice = devices[0];
+        Device foundDevice = config.findDevice(originalDevice.getDeviceName());
+
+        Assert.assertTrue(deepCompare(foundDevice, listedDevice));
```
craigtmoore, line 132: "Found device does not match listedDevices"

dcm4che-conf/dcm4che-conf-dicom/src/test/java/org/dcm4che3/conf/dicom/DicomConfigurationTest.java
```diff
@@ -91,6 +93,88 @@ public void renameAETest() throws ConfigurationException {
...
+    }
+
+    @Ignore
```
craigtmoore, line 135: Why is this on `@Ignore`? I'd just remove it if you're not going to run it.

craigtmoore (approved):
Please see my comments and address them if you can. @stephenfrederick please review this as well, it is something that falls in your team's area of interest.

dcm4che-conf/dcm4che-conf-dicom/src/test/java/org/dcm4che3/conf/dicom/DicomConfigurationTest.java
RhuanKrum: I initially thought about keeping this test method as a reference. But I'll remove it since it doesn't seem to provide any value.
I'll also improve the assertion messages for the other methods I had created

dcm4che-conf/dcm4che-conf-core/src/main/java/org/dcm4che3/conf/core/storage/InMemoryReadOnlyConfiguration.java
```diff
@@ -69,6 +69,13 @@ public Object getConfigurationNode(Path path, Class configurableClass) throws Co
         return Nodes.getNode(root, path.getPathItems());
     }
 
+    @Override
+    public List<Object> getConfigurationNodes(Class configurableClass, Path... paths) throws ConfigurationException {
```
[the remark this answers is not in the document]
RhuanKrum: Pushed e03d3b7a45c3b937056d49b27d3614fe0d34c30a

dcm4che-conf/dcm4che-conf-dicom/src/test/java/org/dcm4che3/conf/dicom/DicomConfigurationTest.java
RhuanKrum: Pushed 8bccf310095393fbf90ab1291b00c381d251554f to address changes on DicomConfigurationTest

craigtmoore (approved):

hczedik (approved):
(I have not reviewed this in detail, as changes in the configuration will not affect the client.)

stephenfrederick (approved):
For future pull requests, you should avoid including the JIRA reference.

RhuanKrum:
There are use cases where we might benefit from having the new `DicomConfiguration#listDevices` to expect multiple or empty arguments (all the devices will be retrieved in this case). 
I just pushed a new commit to address this opportunity.

dcm4che-conf/dcm4che-conf-dicom/src/main/java/org/dcm4che3/conf/dicom/CommonDicomConfiguration.java
```diff
@@ -269,6 +270,41 @@ public Device findDeviceByUUID(String uuid) throws ConfigurationException {
         }
     }
 
+    @Override
+    public Device[] listDevices(DeviceType... primaryDeviceTypes) throws ConfigurationException {
+        if (primaryDeviceTypes == null) {
+            throw new IllegalArgumentException("Requested primaryDeviceTypes cannot be null");
```
craigtmoore, line 276: I don't think this should be here. I thought that this method could cope with a null primaryDeviceTypes (see interface declaration javadoc). The listDeviceNamesBy(primaryDeviceTypes) will return an empty list if primaryDeviceTypes is null (line 281).

craigtmoore (approved):

dcm4che-conf/dcm4che-conf-dicom/src/main/java/org/dcm4che3/conf/dicom/CommonDicomConfiguration.java
RhuanKrum: Now that you mentioned, I acknowledge this can be misleading :anguished:

_`primaryDeviceTypes Can be optionally included. If not, all the {@link Device}s will be retrieved`_
My intention here was to allow a value to either be included in the call or not, as below:
```
configuration.listDevices(); // We're passing an empty array here
configuration.listDevices(DeviceType.PRINT);
configuration.listDevices(DeviceType.PRINT, DeviceType.ARCHIVE);
```
The `IllegalArgumentException` would be thrown if the method is called by explicitly passing `null`:
```
configuration.listDevices(null);
```

When I pushed my latest changes I thought it would be risky allowing null arrays to be passed on, as we could end up having unintended results being retrieved. But this risk actually involves empty arrays the same way. :boom:

One thing I could do to make things safer and easier to use is to split this method according to the acceptable parameters, explicitly (no arguments, one DeviceType, multiple DeviceTypes). This way the caller would know exactly what to expect from each method:
```
Device[] listDevices(); // Will list all devices
Device[] listDevices(DeviceType primaryDeviceType); // List devices according to DeviceType - Null returns an empty Device array
Device[] listDevices(DeviceType[] primaryDeviceTypes); // List devices according to DeviceTypes - Null or empty array returns an empty Device array
```

Would you think this makes more sense?

craigtmoore: Yeah, I think overloading the method makes the more sense, though maybe the first method could be called listAllDevices(), just to be clear.

RhuanKrum: Pushed c7ea96d to address these changes

hczedik (approved):

stephenfrederick (approved):

greg-knox (approved):

Issue #783, opened 2020-09-06
The libopencv_java.so (linux-x86) in binary tool doesn't work in Ubuntu ARM environment


welsleywu:
When using the dcm2jpg tool, it shows the libopencv_java.so (linux-x86) does't not mathch current environment, but the environment is scripts auto detect. 
OS: Ubuntu ARM architecture

Following is the error details:

Exception in thread "main" java.lang.UnsatisfiedLinkError: /root/ftimage/dcm4che/lib/linux-x86_64/libopencv_java.so: /root/ftimage/dcm4che/lib/linux-x86_64/libopencv_java.so:

nroduit:
The next release will bring native libraries for ARM: linux-aarch64 and linux-armv7a

welsleywu:
Thanks for your great help, the new arm SO file works fine, but it requires glibc2.27+. Is it possible also supporting lower version glibc2.17?

nroduit:
Here are the dependencies of the linux build:
``` bash
$ ./glibc-check.sh libopencv_java-4.5.0-dcm-linux-aarch64.so 
GLIBC 2.17
GLIBC 2.27

$ ./glibc-check.sh libopencv_java-4.5.0-dcm-linux-armv7a.so 
GLIBC 2.4
GLIBC 2.6
GLIBC 2.7

$ ./glibc-check.sh libopencv_java-4.5.0-dcm-linux-s390x.so 
GLIBC 2.2
GLIBC 2.2.4
GLIBC 2.3
GLIBC 2.3.2
GLIBC 2.3.4
GLIBC 2.4
GLIBC 2.6
GLIBC 2.7

$ ./glibc-check.sh libopencv_java-4.5.0-dcm-linux-x86.so 
GLIBC 2.0
GLIBC 2.1
GLIBC 2.1.3
GLIBC 2.2
GLIBC 2.2.4
GLIBC 2.3
GLIBC 2.3.2
GLIBC 2.3.4
GLIBC 2.4
GLIBC 2.6
GLIBC 2.7
GLIBC 2.11

$ ./glibc-check.sh libopencv_java-4.5.0-dcm-linux-x86-64.so 
GLIBC 2.2.5
GLIBC 2.3
GLIBC 2.3.2
GLIBC 2.3.4
GLIBC 2.4
GLIBC 2.6
GLIBC 2.7
GLIBC 2.11
GLIBC 2.14
```

[...]

PR #825, opened 2020-11-15
Update native library to 4.5.0

nroduit:
- Update to OpenCV 4.5.0 and first fully automated build with github actions and dockcross
- Support new processor architectures #783: linux-aarch64 linux-armv7a and linux-s390x
- Allow using toolkit on systems other than Linux and Windows
- Fix issues reading raw images with bigendian processors (eg linux-s390x) => native part only used in dcm4che6
- Fix encoding signed data with 1.2.840.10008.1.2.4.57 with bitstored less than 16 (Force encoder using 16-bit but keep current bitstored in DICOM)
- Fix RGB conversion from raw YBR_FULL to 1.2.840.10008.1.2.4.80 or 1.2.840.10008.1.2.4.81 (Force converting to RGB color model)
- Adapt transfer syntax when transcoding (to convert correctly all the files with dcm2dcm): 
      - convert 1.2.840.10008.1.2.4.50 to 1.2.840.10008.1.2.4.70 when bitstored is more than 8
      - convert 1.2.840.10008.1.2.4.51, 1.2.840.10008.1.2.4.53, 1.2.840.10008.1.2.4.55 to 1.2.840.10008.1.2.4.70 when bitstored is less than 13 and with signed data
      - uncompress when bitstored is more than 16

Note: only linux-x86-64, macosx-x86-64, windows-x86-64 and windows-x86 are packaged in dcm4che3 toolkit. Other architectures (linux-x86, linux-aarch64, linux-armv7a...) can be download automatically.

### DIFF
--- a/dcm4che-assembly/pom.xml
+++ b/dcm4che-assembly/pom.xml
@@ -503,6 +503,14 @@
       <version>1.2-pre-dr-b04</version>
     </dependency>
     <dependency>
+      <groupId>com.sun.media</groupId>
+      <artifactId>clib_jiio</artifactId>
+      <classifier>osgi</classifier>
+      <type>zip</type>
+      <version>1.2-pre-dr-b04</version>
+      <scope>runtime</scope>
+    </dependency>
+    <dependency>
       <groupId>org.weasis.thirdparty.org.opencv</groupId>
       <artifactId>libopencv_java</artifactId>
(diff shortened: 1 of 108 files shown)


gunterze:
Thx! (Sorry, missed the notification about that PR for some reason).

Issue #813, opened 2020-11-06
Compression of JPEG Lossless images corrupts pixel data when bits allocated is 16 and bits stored is 8


gunterze:
[decompressed.dcm.gz](https://github.com/dcm4che/dcm4che/files/5502159/decompressed.dcm.gz)
```
$ dcm2dcm --jpll decompressed.dcm compressed.dcm
$ dcm2dcm compressed.dcm decompressed2.dcm
$ dcmdump decompressed.dcm > decompressed.dcm.txt
$ dcmdump decompressed2.dcm > decompressed2.dcm.txt
$ diff decompressed.dcm.txt decompressed2.dcm.txt
139c139
< 3012: (7FE0,0010) OW #102400 [3\3\4\4\4\4\4\4\4\4\3\3\4\3\3\3\3\3\4\4\4\2\3\3\
---
> 3012: (7FE0,0010) OW #102400 [3\0\3\0\4\0\4\0\4\0\4\0\4\0\4\0\4\0\4\0\3\0\3\0\
```

nroduit:
I publish a [PR](https://github.com/dcm4che/dcm4che/pull/815)

Issue #741, opened 2020-06-14
Decompression of JPEG Lossless images fails when bits allocated is 16 and bits stored is 8


wtellis:
**Describe the bug**
Decompression of JPEG Lossless (1.2.840.10008.1.2.1) images fail when the bits allocated is 16 and but the bits stored is only 8.  It may be an issue with `Transcoder.decompressFrame` where the image returned by `adjustColorModel(decompressor.read(0, decompressParam))` is only 8 bits instead of 16 bits.  This means writeFrame writes out 8 bit data instead of 16 bit data and the pixel data tag ends up with half the number of bytes its supposed to have, resulting in a corrupt image. 

**To Reproduce**
`dcm2dcm.bat -t 1.2.840.10008.1.2.1 compressed.dcm decompressed.dcm`
(see attached images)

**Expected behavior**
The resulting decompressed image should contain the correct number of bytes in the pixel data tag. 

**Desktop (please complete the following information):**
 - N/A

**Additional context**
(see attached images)

[decompression issue.zip](https://github.com/dcm4che/dcm4che/files/4775894/decompression.issue.zip)

PR #826, opened 2020-11-17
Fix compression/decompression of images with bitstored is 8 and bitsallocated is 16 in legacy compressor/decompressor

rahulgulati1986:
As suggested in conversation https://groups.google.com/u/1/g/dcm4che/c/xi0S-BtO_lY, made the change in legacy compressor/decompressor on issue when bitstored is 8 and bitsallocated is 16


gunterze:
Related: #741, #813

PR #880, opened 2021-01-25
Added startSequence() and endSequence() methods to SequenceVisitor

patrick-peer:
I extended the SequenceVisitor to include the start and end of sequences. This allowed me to significantly simplify my code, and I think this could generally be useful. I marked the new methods as `default` with an empty implementation for backwards compatibility.


gunterze (changes requested):
Just align the indent of the two added methods and move them below the regular interface methods, to follow the convention of OpenJDK, to put default methods not related to an regular interface method below all regular interface methods.

gunterze (approved):

patrick-peer:
Ah yes, tabs vs spaces never gets old, I guess :-). Thanks for the hint about the method order! I was not aware there were any conventions concerning default methods. Good to know.

PR #883, opened 2021-01-28
Runtime dataset extraction 

shermende:
Hi, I apologize for straying from the topic, I want to say thank you for the wonderful library. I would like to suggest a runtime dataset extraction option, as I have encountered such a problem in my work and have not been able to find solutions in dcm4che or dcm4chee-arc-light.

I ask you to consider such a solution, perhaps it will be useful to you, and you will add such support directly to the DicomInputStream class. If you think the solution is viable I can continue PR or you can directly add it to the library.

PS. if you find the solution incorrect, please tell me how best I could extract the dataset in real time with subsequent streaming to the storage, without first writing and then reading to the stop tag.

### DIFF
--- a/dcm4che-tool/dcm4che-tool-storescp/pom.xml
+++ b/dcm4che-tool/dcm4che-tool-storescp/pom.xml
@@ -47,7 +47,15 @@
   <artifactId>dcm4che-tool-storescp</artifactId>
   <name>dcm4che-tool-storescp</name>
   <description>Receive DICOM Composite Objects from peer DICOM AE</description>
+  <properties>
+    <commons-io.version>2.7</commons-io.version>
+  </properties>
   <dependencies>
+    <dependency>
+      <groupId>commons-io</groupId>
+      <artifactId>commons-io</artifactId>
+      <version>${commons-io.version}</version>
+    </dependency>
     <dependency>
       <groupId>org.dcm4che</groupId>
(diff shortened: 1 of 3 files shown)


gunterze:
s. [Transcoder#trancode](https://github.com/dcm4che/dcm4che/blob/master/dcm4che-imageio/src/main/java/org/dcm4che3/imageio/codec/Transcoder.java#L407) used by dcm4chee-arc-light's [StoreServiceImpl#writeToStorage](https://github.com/dcm4che/dcm4chee-arc-light/blob/master/dcm4chee-arc-store/src/main/java/org/dcm4chee/arc/store/impl/StoreServiceImpl.java#L211).

shermende:
Thank you for your quick response and link. I agree with you about the comments. In the case of DicomInputStream#readDatasetAll (), I ran into a performance issue. The solution allowed me to get the dataset up before pixel data and then stream the data in its original form.

gunterze:
You may debug running `dcm2dcm` utility without compressing/decompressing to see how  [Transcoder#trancode](https://github.com/dcm4che/dcm4che/blob/master/dcm4che-imageio/src/main/java/org/dcm4che3/imageio/codec/Transcoder.java#L407) avoids to allocate the Pixel Data in memory.

shermende:
I studied Transcoder in detail and saw that there is a DIS to DOS streaming, thank you very much.

PR #887, opened 2021-02-02
Fix duplicate entries

nroduit:


gunterze (commented):
Sorry, was confused by reference of both commits...

gunterze (approved):

Issue #42, opened 2017-03-20
dcm2jpg throws EOFException when converting deflated explicit VR little endian file


KarlSjostrand:
Platform: macOS Sierra 10.12.3
dcm4che version: 3.3.8

To reproduce: convert file with TS explicit VR little endian to deflated using `dcm2dcm -t 1.2.840.10008.1.2.1.99 file.dcm file_defl.dcm`. Create jpeg file using `dcm2jpg file_defl.dcm file.jpg`

Result:
```
Failed to convert file_defl.dcm: null
java.io.EOFException
	at javax.imageio.stream.ImageInputStreamImpl.readFully(ImageInputStreamImpl.java:353)
	at javax.imageio.stream.ImageInputStreamImpl.readFully(ImageInputStreamImpl.java:373)
	at org.dcm4che3.imageio.plugins.dcm.DicomImageReader.readRaster(DicomImageReader.java:295)
	at org.dcm4che3.imageio.plugins.dcm.DicomImageReader.read(DicomImageReader.java:342)
	at org.dcm4che3.tool.dcm2jpg.Dcm2Jpg.readImage(Dcm2Jpg.java:381)
	at org.dcm4che3.tool.dcm2jpg.Dcm2Jpg.convert(Dcm2Jpg.java:360)
	at org.dcm4che3.tool.dcm2jpg.Dcm2Jpg.mconvert(Dcm2Jpg.java:345)
	at org.dcm4che3.tool.dcm2jpg.Dcm2Jpg.main(Dcm2Jpg.java:314)
```

Does dcm4che's DicomImageReader not support deflated datasets?

william00179:
Just some overdue follow up on this. Per the header comments in the DicomImageReader, we can pass a few different types into the setInput() function, those being an InputStream, DicomMetaData or ImageInputStream. 

> Supports compressed and uncompressed images from a DicomMetaData object, an InputStream/DicomInputStream or an ImageInputStream.
>  * For ImageInputStream, the access supports random/out of order reading from the input for everything except deflated streams.
>  * For InputStream type data, only sequential access to images is supported, including deflated.
>  * For DicomMetaData, random access is fully supported, and can have been read from a deflated stream.

So I suspect you're passing an ImageInputStream which the DicomImageReader is not able to inflate. 

Try pass in a DicomInputStream and it will inflate the InputStream for you.

PR #889, opened 2021-02-03
Use DicomInputStream to allow for deflated transfer syntaxes to work in dcm2jpg tool

william00179:
Currently, the dcm2jpg tool will throw an EOFException #42 when trying to convert a DICOM file to JPEG that is encoded with a deflated ts.

Per [DicomImageReader](https://github.com/dcm4che/dcm4che/blob/master/dcm4che-imageio/src/main/java/org/dcm4che3/imageio/plugins/dcm/DicomImageReader.java) we can provide an ImageInputStream, DicomMetaData or InputStream. Using an ImageInputStream as the tool currently does, does not support deflated files. Simply swapping this to a DicomInputStream allows for the same functionality and works for deflated input streams also. 

If this isn't acceptable it would be possible to check if the file is deflated in the tool and select either an ImageInputStream or DicomInputStream at that time.

### DIFF
--- a/dcm4che-tool/dcm4che-tool-dcm2jpg/README.md
+++ b/dcm4che-tool/dcm4che-tool-dcm2jpg/README.md
@@ -50,6 +50,12 @@ Options:
     --suffix <suffix>         file extension used with destination
                               directory argument,lower case format name by
                               default
+    --usedis                  use DicomInputStream for reading the DICOM
+                              image. Supports deflated transfer syntaxes
+                              (default, without option --frame <number>)
+    --useiis                  use ImageInputStream for reading the DICOM
+                              image (default, with option --frame
+                              <number>)
     --uselut                  use explicit VOI LUT in image, even if the
                               image also specifies Window Center/Width;
                               prefer applying Window Center/Width over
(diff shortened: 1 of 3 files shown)


gunterze (commented):
Using `FileImageInputStream` avoids to have to read all previous frames from the file on rendering a (particularly not the first) frame. So I would still use it if option `--frame <number>` is specified.
Optionally you may make it configurable, e.g. by adding options
```
    --usedis                  use DicomInputStream for reading the DICOM
                              image (default, without option --frame
                              <number>)
    --useiis                  use ImageInputStream for reading the DICOM
                              image (default, with option --frame
                              <number>)
```
or
```
    --input <class>           specifies class (DicomInputStream or
                              ImageInputStream) to use reading the DICOM
                              image(s). By default ImageInputStream is used
                              if --frame <number> is specified, otherwise
                              DicomInputStream.
```

william00179:
Thanks gunterze, I hadn't considered the FileImageInputStream gives random access to the file. I have added the options as you suggested.

gunterze:
Just did some small changes makeing use of lambdas and force mutual exclusion of  --usedis & --useiis

PR #952, opened 2021-05-07
fix: check if the file is of the directory type in class DicomDirWriter

joaojuniormail:


gunterze:
```
DicomDirWriter open(File file)
```
expects the DICOMDIR file, not the directory!

PR #955, opened 2021-05-12
IEI-132036 Improve memory usage for completed instance locators

komaldeeps96:


komaldeeps96:
clearing the attributes for all locators caused failures in audit messages when accessing patient id-issuers.
one locator will have attributes on it as pid-issuer is same for all locators in a list.

alan-farquhar (approved):

homerocda (approved):

greg-knox (approved):

PR #970, opened 2021-06-09
IEI-81066 Set specific version of maven-deploy-plugin

hczedik:
Cherry-pick of 63af1976e5adf820dc02e56310b73dfa82727cd0

@craigtmoore: I realized that the altDeploymentRepository property only works with this new version of the deploy-plugin. Therefore I am fixing it to be on the safe side (sometimes the build slave is picking an older version of said plugin.)
We might have to backport this little change also to rc branches should we need to release them again and the release fails.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -103,6 +103,10 @@
           </configuration>
         </plugin>
         <plugin>
+          <artifactId>maven-deploy-plugin</artifactId>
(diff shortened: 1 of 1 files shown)


craigtmoore:
Ok, I assumed (incorrectly) that it would work. Thanks for following up.

craigtmoore (approved):

PR #978, opened 2021-06-23
IEI-136521 guessTransferSyntax handles DICOM fragments (no preamble) …

mdoris:
…with a value that containing DICM in the 129th byte.

Changes are to first perform the preamble detection and attempted reading of the fmi and if that fails, to then attempt to guessTransferSyntax using what was incorrectly identified as the preamble. If both checks fail, then throw exception NOT A DICOM STREAM as per previous behaviour.

Example test included showing a StationName of 'DICMA1' appearing in bytes 129-134 which puts 'DICM' in bytes 129-132 and triggered the preamble detection in the guessTransferSyntax.


alan-farquhar:
> …with a value that containing DICM in the 129th byte.
> 
> Changes are to first perform the preamble detection and attempted reading of the fmi and if that fails, to first attempt to guessTransferSyntax using what was incorrectly identified as the preamble. If both checks fail, then throw exception NOT A DICOM STREAM as per previous behaviour.
> 
> Example test included showing a StationName of 'DICMA1' appearing in bytes 129-134 which puts 'DICM' in bytes 129-132 and triggered the preamble detection in the guessTransferSyntax.

alan-farquhar:
Please remove IEI-136521 from the commit message...Even though this is our stream as i would like you to later put a PR through to master you should not use internal JIRA issues.

hczedik:
@mdoris Can you open a PR to master with these changes right away? Would be really good to get a review from @gunterze for these changes.

hczedik:
> Please remove IEI-136521 from the commit message...Even though this is our stream as i would like you to later put a PR through to master you should not use internal JIRA issues.

... you anyways need to recommit/cherry-pick to a new branch to be able to merge to master. So you can also modify the commit message there and leave it here as it is.

dcm4che-core/src/main/java/org/dcm4che3/io/DicomInputStream.java
```diff
@@ -825,31 +825,65 @@ private boolean hasZLIBHeader() throws IOException {
...
+        }
+
         mark(132);
-        int rlen = StreamUtils.readAvailable(this, b132, 0, 132);
-        if (rlen == 132) {
-            if (b132[128] == 'D' && b132[129] == 'I' && b132[130] == 'C' && b132[131] == 'M') {
+        hasfmi = TagUtils.isFileMetaInformation(
+                ByteUtils.bytesToTag(fmiBytes, 0, bigEndian));
+    }
+
```
alan-farquhar, line 864: Please add a comment about what this is doing

hczedik:
Marked as Draft, because of the problem with the testdata. Informed Mike in chat.

hczedik:
@mdoris : Gunter has fixed the issue with a simplified commit in https://github.com/dcm4che/dcm4che/pull/982. I think we should apply the same fix here for the generic-config branch. (If it indeed fixes our case reliably).

mdoris:
> @mdoris : Gunter has fixed the issue with a simplified commit in #982 (see [1d21e72](https://github.com/dcm4che/dcm4che/commit/1d21e72458dcb6b4a2010e4f4985450e062dc4fa)). I think we should apply the same fix here for the generic-config branch. (If it indeed fixes our case reliably).

The fix in Gunters change doesn't fix the problem but instead addresses this instance of the problem. The problem stems from the "magic value" detection and assuming that there cannot be any other valid occurrences of the magic value at the magic byte position.

The original fix handles the scenario when the magic value detection fails which this change does not address.

hczedik:
@gunterze, it seems that @mdoris has good reasons to prefer his version of the fix over yours. Your fix indeed again has the same problem with a magic "DICM20" value that could also be contained by coincidence at the byte position and then again would fail.
What are your reasons to have chosen this solution over Mike's?

gunterze:
I estimate the remaining likelihood  of the occurrence of the 6 byte sequence in binary encoded  DICOM Data Sets - particularly that a string value "DICM" is followed by binary 02 00 -  as not high enough to increase the complexity and therefore maintenance costs for me personally. s.o. [A Philosophy of Software Design](https://www.amazon.com/Philosophy-Software-Design-2nd/dp/173210221X/ref=sr_1_1?dchild=1&qid=1628061075&refinements=p_27%3AJohn+Ousterhout&s=books&sr=1-1&text=John+Ousterhout), 8 Pull Complexity  Downwards

mdoris:
The original magic value was expected to not have a high enough chance of occurrence to incorrectly identify the fmi yet that was already proven incorrect. This problem is also occurs in DICOM fragments which makes the ordering/expectation of following tags less predictable than a full dicom instance/file. As this magic value is written, it simply requires a value ending DICM to start at byte 129 and be followed by any tag in the 0002 group.

The additional complexity is 2 array copies and catching/handling an error. All other actions such as reading from the stream and decisions made from the read data is entirely the same as the original.

gunterze:
Give me ONE scenario where ASCII coded string "DICM" may be followed by binary bytes 2 and 0 in a valid ACR/NEMA stream! Binary 2  is not allowed in DICOM string values , and the two bytes after a string value contains the group tag, whereas tags (0002,eeee) are restricted to the File Meta Information in DICOM files . So I can't follow your conclusion, that because "DICM" actually occurs in existing ACR/NEMA streams,  that the likelihood of the occurrence of  "DICM"   followed by binary byes 2 + 0 is not significantly less.

mdoris:
You yourself posited a valid scenario and while you believe that it is a remote chance, should that occur the data would either be incapable of being ingested or be incapable of being returned upon request in the event the application is being used in a VNA capacity.

When the cost of resolving a potential problem is essentially nil as the solution is already developed and the additional 'complexity' is literally an array copy, it seems negligent to allow the possible situation to develop knowing that should it occur the data would be unavailable/unstorable.

Especially, since the two solutions are easily combinable, having the DICM[20] magic value with the fallback would allow for most to all scenarios to be handled as per your current solution and should that ever prove to be insufficient/erroneous then there is already a fallback in place.

Issue #981, opened 2021-06-24
DicomInputStream.guessTransferSyntax cannot guess when there is no preamble and DICM is contained in data at byte 129


mdoris:
When a DicomInputStream is initialized using a DICOM fragment that contains the value DICM in bytes 129-132 the DicomInputStream will fall into the preamble detection when guessing the transfer syntax. This fails as there is not actually a preamble and the following 132 bytes are not guaranteed to be an fmi as expected. The result is that these well formed DICOM fragments will instead result in a NOT A DICOM STREAM exception being thrown.

Attempt to open a DICOM input stream using a fragement (no preamble) that contains a value 'DICM' in the 129-132 bytes in the fragment. See attached.
[no_preamble_dicm_in_data.zip](https://github.com/dcm4che/dcm4che/files/6710020/no_preamble_dicm_in_data.zip)

**Expected behavior**
A DicomInputStream created using the DICOM fragment

PR #982, opened 2021-06-24
guessTransferSyntax handles value (not prefix) with DICM at byte 129

mdoris:
https://github.com/dcm4che/dcm4che/issues/981

DicomInputStream handles creating a stream for a DICOM fragment that does not contain the preamble or DICM prefix, but does contain a value that happens to include DICM at byte 129.

First attempt to read the preamble/fmi as per usual. If that fails, check if the incorrectly identified preamble can be used to guess the transfer syntax.

Includes new test for fragment with DICM at byte 129


hczedik:
Hi @gunterze,
this is a rather urgent issue for us. What do you think about the fix?
We would very much appreciate your review.
But please do NOT merge it yet, as there is a small problem with the used testdata still.

gunterze:
Fixed by https://github.com/dcm4che/dcm4che/commit/1d21e72458dcb6b4a2010e4f4985450e062dc4fa

PR #1004, opened 2021-08-03
Issue 1003 - fix an out of memory exception when decoding a single frame compressed image

wayfarer3130:
The problem occurs because of an assumption about the segmented image input stream being used already having been fully initialized.  That isn't true because the stream was designed to be lazy loaded, which means it contains an end of fragments of MAX_INT.  That caused an int buffer of size MAX_INT to be allocated, which fails all the time.

### DIFF
--- a/dcm4che-imageio-opencv/src/main/java/org/dcm4che3/opencv/StreamSegment.java
+++ b/dcm4che-imageio-opencv/src/main/java/org/dcm4che3/opencv/StreamSegment.java
@@ -116,14 +116,8 @@ public abstract class StreamSegment {
 
     private static StreamSegment getFileStreamSegment(SegmentedInputImageStream iis) {
         try {
-            Class<? extends ImageInputStream> clazz = iis.getClass();
-            Field fStream = clazz.getDeclaredField("stream");
-            Field fCurSegment = clazz.getDeclaredField("curSegment");
-            if (fCurSegment != null && fStream != null) {
-                fCurSegment.setAccessible(true);
-                fStream.setAccessible(true);
 
-                ImageInputStream fstream = (ImageInputStream) fStream.get(iis);
+                ImageInputStream fstream = iis.getStream();
                 Field fRaf = null;
                 if (fstream instanceof FileImageInputStream) {
                     fRaf = FileImageInputStream.class.getDeclaredField("raf");
@@ -133,7 +127,7 @@ public abstract class StreamSegment {
 
                 if (fRaf != null) {
                     fRaf.setAccessible(true);
-                    long[][] seg = getSegments(iis, clazz, fCurSegment);
+                    long[][] seg = getSegments(iis);
                     if (seg != null) {
                         RandomAccessFile raf = (RandomAccessFile) fRaf.get(fstream);
                         /*
@@ -147,7 +141,7 @@ public abstract class StreamSegment {
                     MemoryCacheImageInputStream mstream = (MemoryCacheImageInputStream) fstream;
                     byte[] b = MemoryStreamSegment.getByte(MemoryStreamSegment.getByteArrayInputStream(mstream));
                     if (b != null) {
-                        long[][] seg = getSegments(iis, clazz, fCurSegment);
+                        long[][] seg = getSegments(iis);
                         if (seg != null) {
                             int offset = (int) seg[0][0];
                             return new MemoryStreamSegment(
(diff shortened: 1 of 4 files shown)


dcm4che-imageio-opencv/src/main/java/org/dcm4che3/opencv/StreamSegment.java
```diff
@@ -157,28 +151,22 @@ private static StreamSegment getFileStreamSegment(SegmentedInputImageStream iis)
...
-                flastSegment.setAccessible(true);
-                List<Object> fragments = (List<Object>) ffragments.get(iis);
-                Integer lastSegment = (Integer) flastSegment.get(iis);
+                List<Object> fragments = iis.getFragments();
+                Integer lastSegment = iis.getLastSegment();
                 if (!desc.isMultiframe() && lastSegment < fragments.size()) {
                     lastSegment = fragments.size();
                 }
-
+                System.out.println("lastSegment="+lastSegment+" curSegment="+curSegment);
```
hczedik, line 169: I think this System.out.println should be removed.

dcm4che-imageio-opencv/src/main/java/org/dcm4che3/opencv/StreamSegment.java
```diff
@@ -157,28 +151,22 @@ private static StreamSegment getFileStreamSegment(SegmentedInputImageStream iis)
...
-            }
         } catch (Exception e) {
             LOGGER.error("Building FileStreamSegment from SegmentedInputImageStream", e);
         }
         return null;
     }
 
-    private static long[][] getSegments(SegmentedInputImageStream iis, Class<? extends ImageInputStream> clazz, Field fCurSegment) throws Exception {
-        Integer curSegment = (Integer) fCurSegment.get(iis);
+    private static long[][] getSegments(SegmentedInputImageStream iis) throws Exception {
```
hczedik, line 160: Do we still need to throw the generic "Exception" here? Can we make it more specific?

hczedik:
Please ignore the failing "Sonarcloud Quality Gate" which says that there is 0% test coverage. The coverage calculation will be fixed in https://github.com/dcm4che/dcm4che/pull/1005.

hczedik:
@gunterze, did you on purpose add the commit be1ed9e9a781fe9329fa0eb9c9110202c41674c3 to this branch here?

hczedik:
> @gunterze, did you on purpose add the commit [be1ed9e](https://github.com/dcm4che/dcm4che/commit/be1ed9e9a781fe9329fa0eb9c9110202c41674c3) to this branch here?

I have reverted that commit, as I assume it was an accident, and it was a slightly different version that you committed to master.

dcm4che-imageio-test/src/test/java/org/dcm4che3/imageio/dcm/TestDicomImageReader.java
```diff
@@ -136,7 +137,19 @@ public void testReadUncompressedPostPixelData_fromImageInputStream() throws IOEx
...
         }
     }
-    
+
+    /**
+     * This test previously throw an out of memory exception, shows that the changes work.
+     * @throws IOException
+     */
+    @Test
+    public void testReadUncompressedingleFrame_fromImageInputStream() throws IOException {
```
hczedik, line 146: Typo: "testReadUncompressedingleFrame_fromImageInputStream" -> "testReadUncompressedSingleFrame_fromImageInputStream"

dcm4che-imageio-test/src/test/java/org/dcm4che3/imageio/dcm/TestDicomImageReader.java
```diff
@@ -136,7 +137,15 @@ public void testReadUncompressedPostPixelData_fromImageInputStream() throws IOEx
             testReadPostPixelData(is);
         }
     }
-    
+
+    @Test
+    public void testReadUncompressedSingleFrame_fromImageInputStream() throws IOException {
```
hczedik, line 142: @wayfarer3130, are you sure that this test is actually covering the scenario that you fixed?
I reverted you fixes, ran the test, and it is still succeeding.

(I just noticed this by coincidence, because I am playing around with jacoco code coverage and noticed that this test is not even running through any of the lines that you have modified.)

dcm4che-imageio-test/src/test/java/org/dcm4che3/imageio/dcm/TestDicomImageReader.java
hczedik: I have corrected the name of the test.

gunterze:
> > @gunterze, did you on purpose add the commit [be1ed9e](https://github.com/dcm4che/dcm4che/commit/be1ed9e9a781fe9329fa0eb9c9110202c41674c3) to this branch here?
> 
> I have reverted that commit, as I assume it was an accident, and it was a slightly different version that you committed to master.

Thx. Thought I did not yet had pushed it, when I realized that I committed it on the wrong branch....

dcm4che-imageio-opencv/src/main/java/org/dcm4che3/opencv/StreamSegment.java
```diff
@@ -116,14 +116,8 @@ public static boolean supportsInputStream(Object iis) {
 
     private static StreamSegment getFileStreamSegment(SegmentedInputImageStream iis) {
```
nroduit, line 117: These changes simplify the access to the segments and allow to limit the "add export" retlated to the reflection (mandatory since Java 16).

dcm4che-imageio-test/src/test/java/org/dcm4che3/imageio/dcm/TestDicomImageReader.java
nroduit: Same as @hczedik. This test is working before the changes.
@wayfarer3130  could you provide a way to get an OOM ?

wayfarer3130: Fixed the test - I had been seeing a couple of exceptions in some other code I was working on for my project static-wado, and thought it was the uncompressed case causing the issues, but in fact it was a compressed single frame issue.  Strangely, it does NOT occur with JPEG-LS, not sure what is different, but does occur on JPLY, for which we have a test image already.  Updated that, confirmed the issue fails without the fixes and passes with the fixes.

dcm4che-imageio-test/pom.xml
```diff
@@ -290,6 +290,11 @@
       <version>${project.version}</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.dcm4che</groupId>
+      <artifactId>dcm4che-imageio</artifactId>
+      <version>${project.version}</version>
+    </dependency>
```
hczedik, line 297: I have added this dependency explicitly, because only that way Jacoco generates a correct aggregated code coverage report for the classes in the dcm4che-imageio.
This increases the report code coverage for these changes from 45.5% to 90.9%.
https://sonarcloud.io/dashboard?id=org.dcm4che%3Adcm4che-parent&pullRequest=1004

hczedik (approved):

hczedik:
Ah, sorry, my approval was not meant as a replacement for an approval from @gunterze.
I hope @gunterze is fine with those changes.

gunterze:
I am not really familiar with  `dcm4che-imageio-opencv/src/main/java/org/dcm4che3/opencv/StreamSegment.java`, so if @nroduit does not see any issues with the change, I am also fine with it.

PR #1008, opened 2021-08-09
Fixed overlay handling if raster was not compatible with color model.

fwiller:


gunterze:
Fixed by a8b1ac2d2 - to fix also apply of Overlays on PALETTE COLOR images

PR #980, opened 2021-06-23
IEI-133247 Large log message captured in failed jobs

Dennis9dl:
The change is already in the dcm4che ei/master branch.
Sorry, I accidently committed the change to dcm4che generic-config branch. So I only add one space at the end here just for this pull request.
Original:  "No compatible connection to " + remote + " available on " + aet);
change:   "No compatible connection to " + remote.getAETitle() + " available on " + aet);

PR #1029, opened 2021-09-09
Reduce the logging by findCompatibelConnection in ApplicationEntity

alsirhc:
This is a continuation of the change from pull request #980, to reduce unnecessary information when findCompatibelConnection in ApplicationEntity throws an exception.

### DIFF
--- a/dcm4che-net-audit/src/main/java/org/dcm4che3/net/audit/AuditLogger.java
+++ b/dcm4che-net-audit/src/main/java/org/dcm4che3/net/audit/AuditLogger.java
@@ -1218,7 +1218,8 @@ public class AuditLogger extends DeviceExtension {
                         return activeConnection;
                     }
         throw new IncompatibleConnectionException(
-                "No compatible connection to " + arr + " available on " + this);
+                "No compatible connection to AuditRecordRepository @ Device " + arr.getDevice().getDeviceName()
+                        + " available on AuditLogger @ Device " + getDevice().getDeviceName());
     }
 
 	public static String processID() {
--- a/dcm4che-net-hl7/src/main/java/org/dcm4che3/net/hl7/HL7Application.java
+++ b/dcm4che-net-hl7/src/main/java/org/dcm4che3/net/hl7/HL7Application.java
@@ -294,7 +294,7 @@ public class HL7Application implements Serializable {
                     if (conn.isInstalled() && conn.isCompatible(remoteConn))
                         return new CompatibleConnection(conn, remoteConn);
         throw new IncompatibleConnectionException(
-                "No compatible connection to " + remote + " available on " + this);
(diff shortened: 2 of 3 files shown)


RhuanKrum (approved):

homerocda (approved):

hczedik:
@alsirhc, the same change should also be opened as a PR to go to the master branch.

(By the way, there the typo "findCompatibelConnection" is already corrected to "findCompatibleConnection" (https://github.com/dcm4che/dcm4che/commit/53b05bcd06eae3b8632373485cbe115aaa929a96). We should also apply that commit here for generic-config. The general goal is to keep generic-config and master as much in sync as possible.

alsirhc:
Based on comments in this pull request, the misspelled method findCompatibelConnection AND potentially the same current problem of too much information dumped in the exception are present in HL7Application in dcm4che/net. Commit 53b05bc addresses only the method name change in master for HL7Application.

It looks to me like the following changes should be applied in a separate pull request:
- correct misspelled method findCompatibelConnection in both ApplicationEntity and HL7Application (pull change from master for this one)
- potentially apply changes in this pull request to HL7Application, as the information in the exceptions for findCompatibelConnection there seem to be as broad as here in ApplicationEntity

greg-knox:
That works for me, thanks @alsirhc

greg-knox (approved):

alsirhc:
After further discussion, the following seems to be the most expedient AND clear course of action:

1. This PR stays as-is, correcting the logging only for findCompatibelConnection in ApplicationEntity. Once this has been verified and tested the change can be cherry-picked and pushed up to master
2. Create a separate artifact to correct findCompatibelConnection to findCompatibleConnection in both classes. This will then align things in master and generic-config
3. Create a separate artifact to similarly reduce logging in  HL7Application

hczedik:
I am still missing the same PR for master.

alsirhc:
Closed because this does not follow the wiki in terms of changing in master first.
Will submit PR to master for change, then cherry pick approved changes into generic-config

alsirhc:
Sorry if this is confusing, but here's a chronological list of changes:

1. Initial change to just update the logging in ApplicationEntity.java, method findCompatibelConnection (ugh, that spelling...)
2. Applied cherry-picked commit 666a572d841dfa1e5c00d37f4e12dd8a80f1eeaf from master, which also brought in some unnecessary changes particularly in ApplicationEntity.java - reverted changes for the latter but accepted cherry-pick changes for AuditLogger.java and HL7Application.java
3. Edited ApplicationEntity.java and HL7Application.java so logging is more like what it would be from the cherry pick, adjusting for some variable names different between master and generic-info

I hope that's clear.

hczedik (approved):

Issue #1031, opened 2021-09-14
build of dcm4che-imageio-test fails on Java 16.0.1 with NPE


gunterze:
```
$ cat ~/work/dcm4che/dcm4che-imageio-test/target/surefire-reports/org.dcm4che3.imageio.dcm.TestDicomImageReader.txt
-------------------------------------------------------------------------------
Test set: org.dcm4che3.imageio.dcm.TestDicomImageReader
-------------------------------------------------------------------------------
Tests run: 21, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 0.286 s <<< FAILURE! - in org.dcm4che3.imageio.dcm.TestDicomImageReader
org.dcm4che3.imageio.dcm.TestDicomImageReader.testReadCompressedSingleFrame_fromImageInputStream  Time elapsed: 0.024 s  <<< ERROR!
java.lang.NullPointerException: Cannot invoke "org.dcm4che3.opencv.StreamSegment.getImageDescriptor()" because "seg" is null
	at org.dcm4che3.imageio.dcm.TestDicomImageReader.testReadCompressedSingleFrame_fromImageInputStream(TestDicomImageReader.java:154)
```
```
~/work/dcm4che/dcm4che-imageio-test$ mvn -V test
Apache Maven 3.8.2 (ea98e05a04480131370aa0c110b8c54cf726c06f)
Maven home: /home/gunter/apache-maven-3.8.2
Java version: 16.0.1, vendor: Private Build, runtime: /usr/lib/jvm/java-16-openjdk-amd64
Default locale: en_US, platform encoding: UTF-8
OS name: "linux", version: "5.11.0-34-generic", arch: "amd64", family: "unix"
[TEST-org.dcm4che3.imageio.dcm.TestDicomImageReader.xml.zip](https://github.com/dcm4che/dcm4che/files/7161446/TEST-org.dcm4che3.imageio.dcm.TestDicomImageReader.xml.zip)
[TEST-org.dcm4che3.imageio.dcm.TestDicomImageReader.xml.zip](https://github.com/dcm4che/dcm4che/files/7161449/TEST-org.dcm4che3.imageio.dcm.TestDicomImageReader.xml.zip)
[TEST-org.dcm4che3.imageio.dcm.TestDicomImageReader.xml.zip](https://github.com/dcm4che/dcm4che/files/7161457/TEST-org.dcm4che3.imageio.dcm.TestDicomImageReader.xml.zip)
[...]

PR #1032, opened 2021-09-14
build of dcm4che-imageio-test fails on Java 16.0.1 with NPE #1031

nroduit:
Add JVM parameters for runtime (from java 8 to 17)

It is required from JDK 16 with legacy API (Decompressor, get Raster) but not with Transcode of dcm2dcm.

As dcm4che3 does not use Java modules, it is necessary to add these options to the boot to allow the use of JDK 17.
```
# Ignore by Java 8
-XX:+IgnoreUnrecognizedVMOptions
# Required from Java 9
--add-opens=java.desktop/javax.imageio.stream=ALL-UNNAMED
--add-opens=java.base/java.io=ALL-UNNAMED
```

Note that JDK 17 does not allow to bypass these options anymore, see this [announcement](https://blogs.oracle.com/javamagazine/java-runtime-encapsulation-internals).

### DIFF
--- a/dcm4che-imageio-test/pom.xml
+++ b/dcm4che-imageio-test/pom.xml
@@ -267,7 +267,7 @@
       <plugin>
         <artifactId>maven-surefire-plugin</artifactId>
         <configuration>
-          <argLine>@{argLine} -Djava.library.path=${project.build.directory}/lib/${os-name}-${cpu-name}</argLine>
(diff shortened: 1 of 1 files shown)


hczedik:
(Just a note: The "All checks have failed" message is due to SonarCloud not supporting analysis of "external" pull requests (coming from a fork). That is a pity. They have a plan to fix this problem: https://jira.sonarsource.com/browse/MMF-1371 ).

gunterze:
Thx!

PR #1034, opened 2021-09-15
Make legacy image APIs compatible with JDK 17

nroduit:
JDK 17 does not allow reflective access of private fields anymore, see this [announcement](https://blogs.oracle.com/javamagazine/java-runtime-encapsulation-internals).


gunterze:
Thx!

PR #1070, opened 2021-12-05
IEI-169967 Add storage version awareness to root configurable classes

msiemczyk:
- Added StorageVersionedConfigurableClass base class that stores the storage version and extended Device, TCConfiguratio, and ConfigurationMetadata with it.
- Modified Reflective Adapter to propagate storage version back and forth between serialized and de-serialized objects. Also updated the schema code to include the storage version as new read-only property and changed Optimistic Lock Hash property to read-only.
- Extracted Optimistic Lock Hash comparison and merge code (not low level algorithm) to new class, called OLockMergeDualFilter, so it can be called outside of olock namespace.
- Added unit tests for new and changed code

### DIFF
--- a/dcm4che-conf/dcm4che-conf-api/src/main/java/org/dcm4che3/conf/api/TCConfiguration.java
+++ b/dcm4che-conf/dcm4che-conf-api/src/main/java/org/dcm4che3/conf/api/TCConfiguration.java
@@ -3,6 +3,7 @@ package org.dcm4che3.conf.api;
 import org.dcm4che3.conf.core.api.ConfigurableClass;
 import org.dcm4che3.conf.core.api.ConfigurableProperty;
 import org.dcm4che3.conf.core.api.ConfigurationException;
+import org.dcm4che3.conf.core.api.StorageVersionedConfigurableClass;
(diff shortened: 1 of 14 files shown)


msiemczyk:
In terms of the generated GUI with my schema changes here is end result for something like device:

image

hczedik:
The changes in "dcm4che-conf" I don't care about, because they are not on master and the client will not use those classes. (Should be pulled in-house.)

The only change that I am not too happy about is the small change in the "Device" class. Instead of further extending that, we should rather bring it inline with the master branch, and pull out all the configuration related stuff into a separate class.

hczedik (approved):

msiemczyk:
@hczedik I know about Device, but this was better design to pass storage version compared to the alternative approach I had so it had to be done. I've attempted to move "dcm4che-conf" to internal repo before and found later on how dcm4che-net with Device depends on on library with core annotations.

I agree eventually we have to do the proteus 2.0 changes of separating Device from config layer, but those are big changes :(

greg-knox (approved):

homerocda (approved):

PR #1076, opened 2021-12-14
IEI-170713 Allows ReferenceIndexingDecorator to work around nodes deleted during refreshes

homerocda:
Calls to ReferenceIndexingDecorator.refreshNode can fail if the backend storage was updated behind the scenes (e.g., when a node updates the database directly without updating the cache first) because it tries to read the uuid of a null node. This change makes sure that this decorator verifies if the node exists before trying to reinsert it into the index.

### DIFF
--- a/dcm4che-conf/dcm4che-conf-core/src/test/java/org/dcm4che3/conf/core/refindexing/RefIndexingTest.java
+++ b/dcm4che-conf/dcm4che-conf-core/src/test/java/org/dcm4che3/conf/core/refindexing/RefIndexingTest.java
@@ -1,34 +1,33 @@
 package org.dcm4che3.conf.core.refindexing;
 
-import org.apache.commons.jxpath.JXPathContext;
 import org.dcm4che3.conf.core.DefaultBeanVitalizer;
 import org.dcm4che3.conf.core.api.Configuration;
 import org.dcm4che3.conf.core.api.Path;
 import org.dcm4che3.conf.core.storage.InMemoryConfiguration;
 import org.dcm4che3.conf.core.index.ReferenceIndexingDecorator;
-import org.dcm4che3.conf.core.util.PathPattern;
 import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Test;
 
-import java.util.Arrays;
 import java.util.HashMap;
-import java.util.List;
 import java.util.Map;
+
+import static org.hamcrest.CoreMatchers.anyOf;
+import static org.hamcrest.CoreMatchers.equalTo;
+import static org.hamcrest.CoreMatchers.hasItem;
+import static org.hamcrest.CoreMatchers.hasItems;
+import static org.hamcrest.CoreMatchers.not;
+import static org.junit.Assert.assertThat;
 
 public class RefIndexingTest {
 
-    private HashMap<String, Path> uuidToSimplePathCache;
-    private Configuration configuration;
-    private DefaultBeanVitalizer vitalizer;
-
-    @Before
-    public void init() {
-        configuration = new InMemoryConfiguration();
-        uuidToSimplePathCache = new HashMap<String, Path>();
-        configuration = new ReferenceIndexingDecorator(configuration, uuidToSimplePathCache);
-        vitalizer = new DefaultBeanVitalizer();
-    }
+    private final HashMap<String, Path> uuidToSimplePathCache = new HashMap<>();
+    private final Configuration lowerLevelConfiguration = new InMemoryConfiguration();
+    private final DefaultBeanVitalizer vitalizer = new DefaultBeanVitalizer();
+    private final Configuration configuration = new ReferenceIndexingDecorator(
+            new SampleCachingConfigDecorator(lowerLevelConfiguration),
+            uuidToSimplePathCache
+    );
 
     @Test
     public void testSimpleChanges() {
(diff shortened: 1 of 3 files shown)


greg-knox (approved):

dcm4che-conf/dcm4che-conf-core/src/test/java/org/dcm4che3/conf/core/refindexing/RefIndexingTest.java
```diff
@@ -11,22 +10,22 @@
...
     private Configuration configuration;
+    private Configuration lowerLevelConfiguration;
     private DefaultBeanVitalizer vitalizer;
 
     @Before
     public void init() {
-        configuration = new InMemoryConfiguration();
-        uuidToSimplePathCache = new HashMap<String, Path>();
-        configuration = new ReferenceIndexingDecorator(configuration, uuidToSimplePathCache);
+        lowerLevelConfiguration = new InMemoryConfiguration();
```
msiemczyk, line 25: Why not just initialize at the class member level? @Before is setup before each test and is the same as class level because JUnit creates new class for each test.

msiemczyk (approved):

dcm4che-conf/dcm4che-conf-core/src/test/java/org/dcm4che3/conf/core/refindexing/RefIndexingTest.java
```diff
@@ -65,6 +64,45 @@ public void testSimpleChanges() {
...
+        sampleBigConf.childList.add(uuid2obj);
+        sampleBigConf.childList.add(new SampleReferableConfClass("UUID3"));
+
+        sampleBigConf.childMap.put("first", new SampleReferableConfClass("UUID4"));
+        sampleBigConf.childMap.put("second", new SampleReferableConfClass("UUID5"));
+
+        configuration.persistNode(new Path("confRoot"), vitalizer.createConfigNodeFromInstance(sampleBigConf), null);
+
+        // check if uuids are indexed
+        Assert.assertEquals(5, uuidToSimplePathCache.size());
```
msiemczyk, line 86: Any chance you can use assertThat and messages for this and asserts on lines 90+?

dcm4che-conf/dcm4che-conf-core/src/test/java/org/dcm4che3/conf/core/refindexing/RefIndexingTest.java
```diff
@@ -65,6 +64,45 @@ public void testSimpleChanges() {
...
+        Assert.assertEquals("Romeo", ((Map) configuration.getConfigurationNode(configuration.getPathByUUID("UUID1"), null)).get("myName"));
+        Assert.assertEquals("Juliet", ((Map) configuration.getConfigurationNode(configuration.getPathByUUID("UUID2"), null)).get("myName"));
+
+        Path pathToRemove = new Path("confRoot", "childMap");
+        // cache the node first
+        configuration.getConfigurationNode(pathToRemove, null);
+        lowerLevelConfiguration.removeNode(pathToRemove);
+        configuration.refreshNode(pathToRemove);
+
+        Assert.assertFalse(uuidToSimplePathCache.containsKey("UUID4"));
```
msiemczyk, line 99: You should use allOf and hasKey Matchers together with assertThat to do this as one assert,

dcm4che-conf/dcm4che-conf-core/src/test/java/org/dcm4che3/conf/core/refindexing/SampleCachingConfigDecorator.java
```diff
@@ -0,0 +1,51 @@
...
+
+    private void putNodeIntoCache(Path nodePath, Object node) {
+        String key = nodePath.toSimpleEscapedPath();
+        if (node != null) {
+            cache.put(key, node);
+        } else {
+            cache.remove(key);
+        }
+    }
+}
```
msiemczyk, line 51: It seems to complain about no end of line here so you should fix that.

dcm4che-conf/dcm4che-conf-core/src/test/java/org/dcm4che3/conf/core/refindexing/RefIndexingTest.java
homerocda: Done

dcm4che-conf/dcm4che-conf-core/src/test/java/org/dcm4che3/conf/core/refindexing/RefIndexingTest.java
homerocda: Done

dcm4che-conf/dcm4che-conf-core/src/test/java/org/dcm4che3/conf/core/refindexing/SampleCachingConfigDecorator.java
homerocda: done

dcm4che-conf/dcm4che-conf-core/src/test/java/org/dcm4che3/conf/core/refindexing/RefIndexingTest.java
homerocda: Done

Issue #1077, opened 2021-12-14
opencv library uses illegal reflection operations in later Java versions


wayfarer3130:
In JDK >= 11, the opencv library gets error messages indicating they are using an illegal reflective operation, which will be going away soon as being allowed.

nroduit:
Yes, this is the current integration, but the future implementation with OpenCV does not contain any more reflections and has a simpler and more direct API.

The problem described here has already been [fixed in the dcm4che commands](https://github.com/dcm4che/dcm4che/pull/1034). For integrations in a third-party application, you have to do the same thing.

wayfarer3130:
@nroduit  - That change is at best a work around - you are hiding the illegal reflection.  Having a clean future implementation without reflections would be good, but in the meantime it is cleaner to have an integration that doesn't rely on ignore warning commands because there are implementations which don't actually use FileImageInputStream or FileCacheImageInputStream.

One way to do that is as this change proposes, to add a method to access the raw underlying data.  That method also can work with other implementations of the image input stream, as long as there actually is a local file somewhere containing the data.

PR #1078, opened 2021-12-14
#1077 Add a get method to get the previously reflected method retrieves

wayfarer3130:
If you run dcm4che imageio opencv decompression under JDK >=11, you get a lot of error messages about reflection becoming illegal shortly and being disallowed.  The changes attached fix this as long as you implement the appropriate methods to get the path name and/or random access file object on the response.

### DIFF
--- a/dcm4che-core/src/main/java/org/dcm4che3/io/PathProvider.java
+++ b/dcm4che-core/src/main/java/org/dcm4che3/io/PathProvider.java
@@ -1,0 +1,5 @@
+package org.dcm4che3.io;
+
+public interface PathProvider {
+    String getPath();
+}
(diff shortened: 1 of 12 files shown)


nroduit:
See my [comment](https://github.com/dcm4che/dcm4che/issues/1077#issuecomment-997679861)

wayfarer3130:
@gunterze - what is your opinion on the change here.  I can add supporting unit tests to make the quality gate pass, but I mean on the underlying issue.  My thought is that the code shouldn't rely on special java allow internal access flags to enable it to work,  but should work on public methods, which is what I modified the code to do.  
@nroduit  - I know you said that there is a longer term change to just be able to direct read from the IIS, but in the meantime, how do you feel about having the extra implementations to allow it to work without the forbidden reflection.

PR #1080, opened 2021-12-15
IEI-169967 Missed updating storage version in reconfigure method

msiemczyk:
- Fixed the reconfigure code and updated the unit test.


dcm4che-net/src/main/java/org/dcm4che3/net/Device.java
```diff
@@ -1281,6 +1281,7 @@ public void reconfigure(Device from) throws IOException, GeneralSecurityExceptio
 
     protected void setDeviceAttributes(Device from) {
         setOlockHash(from.olockHash);
+        setStorageVersion(from.getStorageVersion());
```
msiemczyk, line 1284: I couldn't used private storageVersion member here like they got away for other members because that field is in the base class.

homerocda (approved):

greg-knox (approved):

Issue #1103, opened 2022-01-16
File collisions when storing the same instance twice


or150:
**Describe the bug**
Hi,
I'm using dcm4che storescp in several high load windows servers.
In some scenarios the same SOP Instance is sent to the server from two or more sources at the same time.
This causes several random errors that happen on a daily basis:
* M-RENAME failures in log
* IOExceptions are returned to the sender
* Partial dicoms are being moved to the destination folder

**To Reproduce**
1. Run storescp on a windows server.
2. Send the same instance multiple times in parallel to the server (see script below).
4. The error will happen after

**Expected behavior**
No errors when sending the same instance twice.

**Version Information:**
 - OS: Windows
 - Version 5.24.1

**Additional context**
Python script I used to simulate the error:
```python
import os
import random
import subprocess
import time
from concurrent import futures
from concurrent.futures.thread import ThreadPoolExecutor

MAX_PARALLEL_STUDIES = 10
STUDIES_PER_HOUR = 3000.0*1000.0
SECONDS_IN_HOUR = 3600.0
STUDIES_PER_SECOND = STUDIES_PER_HOUR / SECONDS_IN_HOUR
TOLERANCE_SEC = 30

MAX_STUDIES_TO_SEND = 1000
STORESCP_TARGETS = ['store@127.0.0.1:4243']
DCM4CHE_PATH = os.path.expanduser(r'C:\dcm4che\bin')

pool_executor = ThreadPoolExecutor(max_workers=MAX_PARALLEL_STUDIES)


def send_study(study_path: str, storescp_target: str):
    subprocess.run([os.path.join(DCM4CHE_PATH, 'storescu'), "-c", storescp_target, study_path], check=False,
                   stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL)
    print(f'Sent: {study_path}')


def get_storescp_target():
    while True:
        for target in STORESCP_TARGETS:
            yield target


def get_studies_to_send():
    for _ in range(MAX_STUDIES_TO_SEND):
        yield "D:\Data\one_instance"


def main():
    start_time = time.time()
    end_time = time.time()

    studies_to_send = list(get_studies_to_send())
    random.shuffle(studies_to_send)
    sent_counter = 0
    sending = []
    for study_path, storescp_target in zip(studies_to_send, get_storescp_target()):
        future = pool_executor.submit(send_study, study_path, storescp_target)
        sending.append(future)
[...]

Issue #384, opened 2018-12-11
Support random parts in strings generated by AttributesFormat


gunterze:
Support _Format Element_ `{rnd}`  with optional _Format Type_ `uid` or `uuid` in patterns for `AttributesFormat`.

_Format Element_ | replaced by | sample value
-- | -- | --
`{rnd}` | random 8 hex digits | `93CACC3A`
`{rnd,uid}` | random UID | `2.25.319066559083127606824375875921163553214`
`{rnd,uuid}` | random UUID | `39403141-f5a5-403e-aaef-019576df5c39`

PR #1104, opened 2022-01-16
Feature/handle file collisions

or150:
This fixed issue #1103 for me.

The changes I made:
* Give each temp file a uuid suffix to avoid collision of temp files.
* Use java.nio for move operation to get a more detailed error in logs.
* Retry moves when failed. Added some random jitter to improve success rate. (Wanted to go with ATOMIC_MOVE, but saw too many edge cases in its [definition](https://docs.oracle.com/javase/7/docs/api/java/nio/file/Files.html#move(java.nio.file.Path,%20java.nio.file.Path,%20java.nio.file.CopyOption...))).
* Added new arguments to allow controlling retires.


gunterze:
Specify --filepath <pattern> including {rnd} which will be replaced by random 8 hex digits (#384)

PR #1115, opened 2022-01-25
Blocked dimse timeouts

mdoris:
The first request to abort an association will await the monitor to write the AAbort. The association once aborted will take as the last piece of new data to write the aabort. Any additional threads attempting to write to an aborted association will neither write nor block.


hczedik (approved):
Looks good to me, but I am not familiar with this code part.

Please do also open a PR for master! Would be good to get feedback also from @gunterze as soon as possible.

greg-knox (approved):

PR #1119, opened 2022-01-31
Allow to set total stream length for DicomInputStream (IEI-106938)

hczedik:
Can be used as an alternative to setAllocateLimit() if the stream length is already known upfront.

### DIFF
--- a/dcm4che-core/src/main/java/org/dcm4che3/io/DicomInputStream.java
+++ b/dcm4che-core/src/main/java/org/dcm4che3/io/DicomInputStream.java
@@ -136,7 +136,7 @@ public class DicomInputStream extends FilterInputStream
     }
 
     public DicomInputStream(InputStream in, int preambleLength) throws IOException {
-        super(in.markSupported() ? in : new BufferedInputStream(in));
+        super(ensureMarkSupported(in));
         guessTransferSyntax(preambleLength);
     }
 
@@ -149,6 +149,54 @@ public class DicomInputStream extends FilterInputStream
             throw e;
         }
         uri = file.toURI().toString();
+    }
+
+    /**
+     * Create a new DicomInputStream with the given input stream and read limit.
+     * It ensures to never read more than the limit from the stream by wrapping it with a {@link LimitedInputStream}.
+     *
+     * The limit also helps to avoid OutOfMemory errors on parsing corrupt DICOM streams without the need to create
+     * temporary arrays when allocating large tag values. (See also {@link #setAllocateLimit}.)
+     *
+     * @param in input stream to read data from
+     * @param limit limit in bytes
+     * @return new DicomInputStream
+     * @throws IOException if there is a problem reading from the given stream
+     */
+    public static DicomInputStream createWithLimit(InputStream in, long limit) throws IOException {
+        return new DicomInputStream(new LimitedInputStream(ensureMarkSupported(in), limit, true));
+    }
+
+    /**
+     * Create a new DicomInputStream for the given file.
+     *
+     * A limit will be set by reading the length of the file (see also #createWithLimit).
+     *
+     * @param file file to read
+     * @return new DicomInputStream
+     * @throws IOException if there is a problem reading from the given file
+     */
+    public static DicomInputStream createWithLimitFromFileLength(File file) throws IOException {
+        long fileLength = file.length();
+        // Some operating systems may return 0 length for pathnames denoting system-dependent entities such as devices or pipes
+        if(fileLength > 0) {
+            InputStream in = new LimitedInputStream(new BufferedInputStream(new FileInputStream(file)), fileLength, true);
+            DicomInputStream dicomInputStream;
+            try {
+                dicomInputStream = new DicomInputStream(in);
+            } catch (IOException e) {
+                SafeClose.close(in);
+                throw e;
+            }
+            dicomInputStream.setURI(file.toURI().toString());
+            return dicomInputStream;
+        } else {
+            return new DicomInputStream(file);
+        }
+    }
+
+    private static InputStream ensureMarkSupported(InputStream in) {
+        return in.markSupported() ? in : new BufferedInputStream(in);
     }
 
     public final String getTransferSyntax() {
@@ -181,9 +229,13 @@ public class DicomInputStream extends FilterInputStream
      * OutOfMemoryErrors on parsing corrupted DICOM streams.
      * 
      * By default, the limit is set to 67108864 (64 MiB).
+     *
+     * Note: If a limit is given using {@link #createWithLimit} or
+     * {@link #createWithLimitFromFileLength} or by supplying a {@link LimitedInputStream},
+     * then this allocateLimit will be ignored (except for deflated data) and no
+     * temporary arrays need to be created.
      * 
      * @param allocateLimit limit of initial allocated memory or -1 for no limit
-     * 
      */
     public final void setAllocateLimit(int allocateLimit) {
         this.allocateLimit = allocateLimit;
--- a/dcm4che-core/src/main/java/org/dcm4che3/io/DicomInputStream.java
+++ b/dcm4che-core/src/main/java/org/dcm4che3/io/DicomInputStream.java
@@ -262,6 +262,9 @@ public class DicomInputStream extends FilterInputStream
      * @param allocateLimit limit of initial allocated memory or -1 for no limit
      */
     public final void setAllocateLimit(int allocateLimit) {
+        if(allocateLimit < -1 || allocateLimit == 0)
+            throw new IllegalArgumentException("allocateLimit must be a positive number or -1");
+
         this.allocateLimit = allocateLimit;
     }
 
@@ -967,14 +970,14 @@ public class DicomInputStream extends FilterInputStream
                         "Length " + valLen + " for tag " + TagUtils.toString(tag) + " @ " + tagPos  +
                                 " exceeds remaining " + ((LimitedInputStream)in).getRemaining() +  " (pos: " + pos + ")");
             }
-            int allocLen = allocateLimit >= 0 && !limitedStream
+            int allocLen = allocateLimit != -1 && !limitedStream
                     ? Math.min(valLen, allocateLimit)
                     : valLen;
             byte[] value = new byte[allocLen];
             readFully(value, 0, allocLen);
             while (allocLen < valLen) {
                 int newLength = allocLen << 1;
-                if (newLength < 0)
+                if (newLength <= 0)
                     newLength = Integer.MAX_VALUE;
                 if (newLength > valLen)
                     newLength = valLen;
--- a/dcm4che-core/src/main/java/org/dcm4che3/io/DicomInputStream.java
+++ b/dcm4che-core/src/main/java/org/dcm4che3/io/DicomInputStream.java
@@ -87,6 +87,9 @@ public class DicomInputStream extends FilterInputStream
     private static final int ZLIB_HEADER = 0x789c;
     private static final int DEF_ALLOCATE_LIMIT = 0x4000000; // 64MiB
 
+    private static final int DEFAULT_PREAMBLE_LENGTH = 128;
+    private static final int UNDEFINED_LENGTH = -1;
+
     // Length of the buffer used for readFully(short[], int, int)
     private static final int BYTE_BUF_LENGTH = 8192;
 
@@ -132,7 +135,7 @@ public class DicomInputStream extends FilterInputStream
     }
 
     public DicomInputStream(InputStream in) throws IOException {
-        this(in, 128);
+        this(in, DEFAULT_PREAMBLE_LENGTH);
     }
 
     public DicomInputStream(InputStream in, int preambleLength) throws IOException {
@@ -143,7 +146,7 @@ public class DicomInputStream extends FilterInputStream
     public DicomInputStream(File file) throws IOException {
         super(new BufferedInputStream(new FileInputStream(file)));
         try {
-            guessTransferSyntax(128);
+            guessTransferSyntax(DEFAULT_PREAMBLE_LENGTH);
         } catch (IOException e) {
             SafeClose.close(in);
             throw e;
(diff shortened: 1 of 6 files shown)


hczedik:
Note: We could also set the stream length in the DicomInputStream(File file) constructor by calling `file.length()`.

hczedik:
Hi @gunterze,

we would appreciate a bit of feedback on this one. Is the general approach okay?
(Just for us to decide how we continue with the issue.)

gunterze:
Why not using existing `int len` parameter of  
```java
    public Attributes readDataset(int len, Predicate<DicomInputStream> stopPredicate) throws IOException {
```
?

hczedik:
I was thinking about this, but it seemed harder to implement, as you would need to pass len through a lot of methods. Through the DicomInputHandler, and also into readFileMetaInformation(), as we also want to consider it there.

Unless you mean that we should save the current value to an instance field. Then it would be pretty much the same as the streamLength field I introduced.

Another concern is that len is an int and not a long.

I am very open to change the implementation, if you give some guidance.

hczedik:
Also, `len` currently is not really a hard limit: It would still read a tag value completely, even if the given len ends somewhere in middle of a value. So if we use it for that the purpose of length checks, then that would be a change in behavior.

hczedik:
By the way, on this line `len` is used for the decision whether to set `includeBulkDataURI` to true.

https://github.com/dcm4che/dcm4che/blob/master/dcm4che-core/src/main/java/org/dcm4che3/io/DicomInputStream.java#L579

Is that maybe a typo? Did you mean `length != 0`?

gunterze:
Would prefer to widen param len from int to long and also use it to validate read/alloc values.

> https://github.com/dcm4che/dcm4che/blob/master/dcm4che-core/src/main/java/org/dcm4che3/io/DicomInputStream.java#L579

is a typo.

hczedik:
Okay, I will try to use the `len` parameter.
But I have to admit that I find that one a bit confusing:
* It only gets checked before reading a tag, so it is a "soft" limit. But now we are making it a harder limit. But only hard in certain cases, other cases will still happily read more than the specified `len`. Seems quite inconsistent to me.
* len gets ignored in the readFileMetaInformation() call. The readAttributes() call later does not consider that fmi was already read, and does not substract that from the len. Maybe it should? But that is also a breaking change to the current behavior.

hczedik:
Okay, that did break a few tests...
I'll look into that tomorrow.

But is this the general idea, how you would like to have it?

hczedik:
TODO: Handle deflated case (when using InflaterInputStream).

hczedik:
Hi @gunterze,

using the len parameter is somehow quite difficult/confusing. Already before my changes, in fact.

I was assuming I could use it like this:
```
File file = new File(...);
try ( DicomInputStream in = new DicomInputStream(file)) {
   in.readDataset(file.length());
}
```

But, no, that doesn't work, because the constructor is already (usually) reading the first 132 bytes.

So I have to do something like this:

```
File file = new File(...);
try ( DicomInputStream in = new DicomInputStream(file)) {
   in.readDataset(file.length() - in.getPosition());
}
```

But, no, that will also not work, as readDataset() might also read the fileMetaInformation and then the given length is again wrong, causing an EOFException.

So, a user could solve this by explicitly reading the fmi before:

```
File file = new File(...);
try ( DicomInputStream in = new DicomInputStream(file)) {
   in.readFileMetaInformation();
   in.readDataset(file.length() - in.getPosition());
}
```

Or, we could fix the readDataset method to correct the length for the case that fmi has not been read yet.
That would be a slight change in behavior to the current logic, but I think it would be a good idea.

But even if I do that, the "problem" that the constructor already reads some bytes, is still there, which makes it very hard to use this API for a user.

What are your thoughts?

hczedik:
Maybe we should relax the `len` parameter a bit, so that it is maximum length, and if a length too long is given it does not cause an EOFException.

gunterze:
> Maybe we should relax the `len` parameter a bit, so that it is maximum length, and if a length too long is given it does not cause an EOFException.

Would also like to defer my review because of my current workload...

gunterze:
> By the way, on this line `len` is used for the decision whether to set `includeBulkDataURI` to true.
> 
> https://github.com/dcm4che/dcm4che/blob/master/dcm4che-core/src/main/java/org/dcm4che3/io/DicomInputStream.java#L579
> 
> Is that maybe a typo? Did you mean `length != 0`?

https://github.com/dcm4che/dcm4che/issues/1122

gunterze:
> Maybe we should relax the `len` parameter a bit, so that it is maximum length, and if a length too long is given it does not cause an EOFException.

Yuu could also always wrap  the passed InputStream by [LimitedInputStream](https://github.com/dcm4che/dcm4che/blob/master/dcm4che-core/src/main/java/org/dcm4che3/util/LimitedInputStream.java)

dcm4che-core/src/main/java/org/dcm4che3/io/DicomInputStream.java
```diff
@@ -484,29 +491,40 @@ public Attributes readCommand() throws IOException {
...
     }
 
-    public Attributes readDataset(int len, Predicate<DicomInputStream> stopPredicate) throws IOException {
+    public Attributes readDataset(long len, Predicate<DicomInputStream> stopPredicate) throws IOException {
         handler.startDataset(this);
+        readDatasetEndPos = (len == UNDEFINED_LENGTH) ? UNDEFINED_LENGTH : pos + len;
+        long posBeforeReadingFMI = pos;
         readFileMetaInformation();
+        if(len != UNDEFINED_LENGTH) {
+            len -= (pos - posBeforeReadingFMI);
```
hczedik, line 520: I'm correcting the `len` here now, in the case fmi was not read yet.

But I am a bit unsure about the whole thing. (See my long previous comment.)

Open points:
* What should we do about the problem with the bytes already read in the constructor. Is it really the user of the API that has to accommodate for that?
* Extend DicomInputStreamTest to cover also reading with a `len` given for different test-data cases.
* Fix the deflated case.

hczedik:
Hi @gunterze,

what do you think about this solution? (committed to a separate branch)
https://github.com/dcm4che/dcm4che/commit/69c3bd4dca45181cedef97222caf9d0f9fa242b6

Thanks for the suggestion to use a LimitedInputStream.
I think in general, that is a nicer approach than using the `len` param of readDataset, as that one has rather confusing behavior (maybe think about deprecating it?)

gunterze:
Vote for

> An alternative would be to not add special constructors, but to rely on the caller to supply the LimitedInputStream, ...

hczedik:
> Vote for
    "An alternative would be to not add special constructors, but to rely on the caller to supply the LimitedInputStream, ..."

Okay... but can I at least add a static helper method, something like `DicomInputStream.createWithLimit(InputStream in, long limit)`?
Otherwise, it is a very hidden feature.

craigtmoore:
I also think there are too many Constructors in https://github.com/dcm4che/dcm4che/commit/69c3bd4dca45181cedef97222caf9d0f9fa242b6. I think would be better to use a builder pattern instead. Otherwise, you can simply rely on the caller to supply the LimitedInputStream as @gunterze  suggests. @gunterze  will the LimitedInputStream also work with compressed dicomInputStreams?

gunterze:
> > Vote for
> > "An alternative would be to not add special constructors, but to rely on the caller to supply the LimitedInputStream, ..."
> 
> Okay... but can I at least add a static helper method, something like `DicomInputStream.createWithLimit(InputStream in, long limit)`? Otherwise, it is a very hidden feature.

I don't care - if you like :)

gunterze:
> I also think there are too many Constructors in [69c3bd4](https://github.com/dcm4che/dcm4che/commit/69c3bd4dca45181cedef97222caf9d0f9fa242b6). I think would be better to use a builder pattern instead. Otherwise, you can simply rely on the caller to supply the LimitedInputStream as @gunterze suggests. @gunterze will the LimitedInputStream also work with compressed dicomInputStreams?

Guess it depends if using (which) ImageReader implementation for decompression.

hczedik:
@gunterze, I've rewritten the history of this PR completely, and nicely split it into separate commits to ease the review.

Please take another look, once you have some time.

Thanks!

dcm4che-core/src/main/java/org/dcm4che3/io/DicomInputStream.java
```diff
@@ -548,11 +548,30 @@ public Attributes readDatasetUntilPixelData() throws IOException {
...
+     * @deprecated Use one of the other {@link #readDataset()} methods instead. If you want to
+     * specify a length limit, you may supply a {@link LimitedInputStream} or use
+     * {@link #createWithLimit} or {@link #createWithLimitFromFileLength}.
+     */
     @Deprecated
     public Attributes readDataset(int len, int stopTag) throws IOException {
         return readDataset(len, tagEqualOrGreater(stopTag));
     }
 
+    public Attributes readDataset(int stopTag) throws IOException {
```
hczedik, line 564: Not sure why the above method was already deprecated before. I actually think using a stopTag can be quite useful, therefore I have added this method here.

hczedik:
@craigtmoore 

>  will the LimitedInputStream also work with compressed dicomInputStreams?

If you mean streams with a deflated transfer syntax, then yes, the implementation I did now handles that nicely: As soon as the LimitedInputStream gets wrapped with an InflaterInputStream, the instanceof check in the readValue method returns false, and it will fall back to the allocateLimit as before.

If you mean other compressions formats and reading images from the stream, then I don't really see how that could be affected by the changes.

dcm4che-core/src/main/java/org/dcm4che3/io/DicomInputStream.java
```diff
@@ -484,22 +561,41 @@ public Attributes readCommand() throws IOException {
...
+        return readDataset(UNDEFINED_LENGTH, stopPredicate);
+    }
+
+    /**
+     * @deprecated Use one of the other {@link #readDataset()} methods instead. If you want to
+     * specify a length limit, you may supply a {@link LimitedInputStream} or use
+     * {@link #createWithLimit} or {@link #createWithLimitFromFileLength}.
+     */
+    @Deprecated
     public Attributes readDataset(int len, Predicate<DicomInputStream> stopPredicate) throws IOException {
```
craigtmoore, line 599: I guess the plan is the eventually make this method private?

craigtmoore:
I've squashed my commits into one and added extra unit tests to improve code coverage. @hczedik, you'll need to rebase your local branch accordingly.

dcm4che-core/src/main/java/org/dcm4che3/io/DicomInputStream.java
hczedik: Or just keep it deprecated like this, as I don't think a library should break user's code, unless it is really needed.

But that is just my suggestion, let's see what @gunterze thinks.

hczedik:
Thanks for adding some unit tests @craigtmoore. But as they are a bit unrelated, they won't increase the coverage of this PR, as Sonarqube is measuring the coverage of modified/new lines.

About the change to ignore an allocateLimit of 0: Probably good to change this, as it would have let to an endless loop, I think. But you also could have considered that as a "user error", as an allocateLimit of 0 clearly does not make any sense.

While at it, you also might want to change the line which checks the newLength to disallow 0:
https://github.com/dcm4che/dcm4che/blob/master/dcm4che-core/src/main/java/org/dcm4che3/io/DicomInputStream.java#L874
`if (newLength < 0)` could also be `if (newLength <= 0)`.
While, with your latest change, the newLength actually will never be 0 I think (does the left-shift operation (<< 1) of any positive number ever result in 0? I don't think so, but I am not 100% sure at the moment.), it still would it make more obvious 0 would by an invalid case here (which would result in an endless loop).

craigtmoore:
OK, I've raised the coverage by a couple of percentage points to make SonarQube happy. @hczedik maybe you can try adding a few more?

hczedik:
I don't think we should only add unit tests to please Sonarqube. The low coverage is due to code that was also not covered before. The new code is already quite nicely covered, in my opinion.

hczedik:
I have now cleaned up the unrelated unit tests again, to have this PR a bit more focused on the actual problem. (@craigtmoore, please feel free to create an extra PR for your extra unit tests.)

The invalid values for allocateLimit I am now rejecting directly in the setter: https://github.com/dcm4che/dcm4che/pull/1119/commits/c4e21ff7b3738a7da0a6efb6f6bdc50591d25280

The Sonarqube ("SonarCloud Code Analysis") complaint about not perfect code coverage, I would just ignore, because that is because of other code that was touched, which was also not covered before.

@gunterze, in my opinion, this PR would be ready for Merge. Sorry for the large changeset all at once, but the commits are nicely separate and by themselves all quite simple. Could you please take another look?
Thanks a lot!

craigtmoore (approved):

dcm4che-core/src/main/java/org/dcm4che3/io/DicomInputStream.java
```diff
@@ -861,10 +957,15 @@ private void readFragments(Attributes attrs, int fragsTag, VR vr)
...
-            int allocLen = allocateLimit >= 0
+            if (valLen < 0) {
+                throw new IOException("internal error: length should have been validated in readHeader");
+            }
+            boolean limitedStream = in instanceof LimitedInputStream;
+            if(limitedStream && valLen > ((LimitedInputStream)in).getRemaining()) {
+                throw new EOFException(String.format("Length %d for tag %s @ %d exceeds remaining %d (pos: %d)",
+                        valLen, TagUtils.toString(tag), tagPos, ((LimitedInputStream) in).getRemaining(), pos));
+            }
+            int allocLen = allocateLimit > 0 && !limitedStream
```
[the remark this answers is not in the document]
hczedik: I know, but I changed it to `allocateLimit != -1` in the latest commits, because now the invalid value of 0 gets already rejected in the setAllocateLimit() setter.

hczedik:
Hi @gunterze,
did you have a chance to take another look at this?
Any feedback if the general approach is fine now?

PR #1126, opened 2022-02-09
Blocked dimse timeouts gc

mdoris:
When multiple timeouts are attempting to block the same association, the first one will queue on the monitor (if needed) while the other timeouts will exit without blocking.


hczedik (approved):

hczedik:
Looks fine to me, approved, but I am not familiar with this code part.

PR #1139, opened 2022-03-10
Cherry pick stream-length changes from master branch to generic-config branch

craigtmoore:
Cherry picking the changes related to verifying the stream length from the master branch to the generic-config branch.

PR #1152, opened 2022-03-24
Revert the revert and add extra if statements to the readAttributes and readItemHeader methods

craigtmoore:
I've reverted the revert to again apply the changes from pull request [#1139](https://github.com/dcm4che/dcm4che/pull/1139). I've also added more details to the warning message when an unexpected attribute or unexpected non-zero length item is encountered. It now includes the method name in the warning message so its easier to figure out which method is generating the message.
```java
    private static final String UNEXPECTED_NON_ZERO_ITEM_LENGTH =
       "Unexpected item value of {} #{} @ {} during {}";
    private static final String UNEXPECTED_ATTRIBUTE =
        "Unexpected attribute {} #{} @ {} during {}";
```

Note the `during {}` added to the end of the string. This combined with changes to the `skipAttribute(String message, String methodName)` where it will now take the `methodName` as a paramter and also log on debug level the number of actual bytes skipped
```java
private void skipAttribute(String message, String methodName) throws IOException {
    String tagAsString = TagUtils.toString(this.tag);
    LOG.warn(message, tagAsString, length, tagPos, methodName);
    long skipLength = skip(length);
    LOG.debug("Skipped {} actual bytes for {}", skipLength, tagAsString);
}
```

I've dropped the changes that cause the cause it to skip logging a warning message when it encounters a SequenceDelimitationItem or ItemDelimitationItem that are zero length.

### DIFF
--- a/dcm4che-core/src/main/java/org/dcm4che3/io/DicomInputStream.java
+++ b/dcm4che-core/src/main/java/org/dcm4che3/io/DicomInputStream.java
@@ -51,6 +51,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
+import java.util.function.Predicate;
 import java.util.zip.Inflater;
 import java.util.zip.InflaterInputStream;
 
@@ -65,6 +66,7 @@ import org.dcm4che3.data.Tag;
 import org.dcm4che3.data.UID;
 import org.dcm4che3.data.VR;
 import org.dcm4che3.util.ByteUtils;
+import org.dcm4che3.util.LimitedInputStream;
 import org.dcm4che3.util.SafeClose;
 import org.dcm4che3.util.StreamUtils;
 import org.dcm4che3.util.TagUtils;
@@ -99,6 +101,9 @@ public class DicomInputStream extends FilterInputStream
 
     private static final int ZLIB_HEADER = 0x789c;
     private static final int DEF_ALLOCATE_LIMIT = 0x4000000; // 64MiB
+
+    private static final int DEFAULT_PREAMBLE_LENGTH = 128;
+    private static final int UNDEFINED_LENGTH = -1;
 
     // Length of the buffer used for readFully(short[], int, int)
     private static final int BYTE_BUF_LENGTH = 8192;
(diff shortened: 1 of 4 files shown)


mdoris (approved):

dcm4che-core/src/main/java/org/dcm4che3/io/DicomInputStream.java
```diff
@@ -708,8 +715,13 @@ public void readAttributes(Attributes attrs, int len, Predicate<DicomInputStream
                     bigEndian = prevBigEndian;
                     explicitVR = prevExplicitVR;
                 }
-            } else
-                skipAttribute(UNEXPECTED_ATTRIBUTE, "readAttributes()");
+            } else if (tag == Tag.SequenceDelimitationItem) {
+                if (length != 0) {
```
hczedik, line 719: I need to see some test data to understand whether this new `if` is an actually valid case.
Also, if this problem is also present on master, where is the PR for master?
A github issue explaining the issue would also be good.

craigtmoore: I've removed this from the pull request. We determined that the test that is failing likely has more problems than this. It is introducing the SequenceDeliminationItem when creates the recreates the header file while storing.

craigtmoore:
I've decided to approach this in a different way because these changes cause too many integration test failures when introduced with our server.

craigtmoore:
@gunterze I've finally got this working properly on the generic-config branch. Can you please have a look at it when you have a chance? It's an attempt to backport @hczedik's changes from the master branch to the generic-config branch.

dcm4che-core/src/main/java/org/dcm4che3/io/DicomInputStream.java
```diff
@@ -83,9 +84,9 @@ public enum IncludeBulkData { NO, YES, URI }
         LoggerFactory.getLogger(DicomInputStream.class);
 
     private static final String UNEXPECTED_NON_ZERO_ITEM_LENGTH =
-        "Unexpected item value of {} #{} @ {}";
+        "Unexpected item value of {} #{} @ {} during {}";
```
hczedik, line 87: I would suggest opening these improvements in the log messages also as a PR for master.

dcm4che-core/src/main/java/org/dcm4che3/io/DicomInputStream.java
```diff
@@ -709,10 +795,16 @@ private void checkIsThis(DicomInputStream dis) {
...
-        skip(length);
+    /**
+     * @param message the message to use in the warning log message
+     * @param methodName the name of the method that is skipping the attribute
+     * @throws IOException potentially thrown when performing the 'skip' operation
+     */
+    private void skipAttribute(String message, String methodName) throws IOException {
+        String tagAsString = TagUtils.toString(this.tag);
+        LOG.warn(message, tagAsString, length, tagPos, methodName);
+        long skipLength = skip(length);
```
hczedik, line 806: Hmm, I wonder if this call to skip(.) here shouldn't actually be a call to skipFully(.)? (@gunterze)

hczedik (approved):

dcm4che-core/src/main/java/org/dcm4che3/io/DicomInputStream.java
craigtmoore: OK, I will create a pull request for this.

dcm4che-core/src/main/java/org/dcm4che3/io/DicomInputStream.java
craigtmoore: I'm going to resolve this conversation so that I can merge my pull request. I don't like this feature of github, you should be able to merge even if there are comments in the pull request. Assuming the reviewer has already approved it.

PR #1155, opened 2022-03-30
Generic config

zliang88:
IEI-187643 The transient field aliasApplicationEntitiesMap in Device class is null in the returned device instances from remote EJB

### DIFF
--- a/dcm4che-net/src/main/java/org/dcm4che3/net/Device.java
+++ b/dcm4che-net/src/main/java/org/dcm4che3/net/Device.java
@@ -48,6 +48,7 @@ import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.List;
@@ -223,7 +224,6 @@ public class Device extends StorageVersionedConfigurableClass implements Seriali
 
     /**
      * Note: This only maps the main AE titles to application entities.
-     * {@link #aliasApplicationEntitiesMap} will contain also alias AE titles.
      */
     @LDAP(noContainerNode = true)
     @ConfigurableProperty(
@@ -240,12 +240,6 @@ public class Device extends StorageVersionedConfigurableClass implements Seriali
             description = "Default AE to be used by both services running locally on this device as well as external services"
     )
     private ApplicationEntity defaultAE;
-
-    /**
-     * Maps alias AE titles ({@link ApplicationEntity#getAETitleAliases()}),
-     * including also the main AE title ({@link ApplicationEntity#getAETitle()}), to application entities.
-     */
-    private final transient Map<String, ApplicationEntity> aliasApplicationEntitiesMap = new TreeMap<String, ApplicationEntity>();
 
     @ConfigurableProperty(name = "deviceExtensions", isExtensionsProperty = true)
     private Map<Class<? extends DeviceExtension>, DeviceExtension> extensions =
@@ -945,7 +939,6 @@ public class Device extends StorageVersionedConfigurableClass implements Seriali
 
     public void setApplicationEntitiesMap(Map<String, ApplicationEntity> applicationEntitiesMap) {
         this.applicationEntitiesMap.clear();
-        this.aliasApplicationEntitiesMap.clear();
         for (Entry<String, ApplicationEntity> entry : applicationEntitiesMap.entrySet()) {
             addApplicationEntity(entry.getValue());
         }
@@ -966,8 +959,6 @@ public class Device extends StorageVersionedConfigurableClass implements Seriali
         ae.setDevice(this);
 
         applicationEntitiesMap.put(ae.getAETitle(), ae);
-
-        addAllAliasesForApplicationEntity(ae);
     }
 
     public ApplicationEntity removeApplicationEntity(ApplicationEntity ae) {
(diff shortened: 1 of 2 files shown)


dcm4che-net/src/main/java/org/dcm4che3/net/Device.java
```diff
@@ -1084,7 +1063,11 @@ public ApplicationEntity getApplicationEntity(String aet) {
      * @return AE titles of this device, including alias AE titles
      */
     public Collection<String> getApplicationAETitles() {
-        return aliasApplicationEntitiesMap.keySet();
+        HashSet<String> aeTitles = new HashSet<>();
```
msiemczyk, line 1066: Nit-pick; usually on the left side you declare it as "Set" interface, rather than concrete class, but not a big deal.

dcm4che-net/src/main/java/org/dcm4che3/net/Device.java
```diff
@@ -1084,7 +1063,11 @@ public ApplicationEntity getApplicationEntity(String aet) {
...
+        aeTitles.addAll(applicationEntitiesMap.keySet());
+        //add alias AE titles
```
msiemczyk, line 1068: Nit-pick: Not sure if the comment adds any value, we can see that by the one liner and the public method comment mentions that too.

dcm4che-net/src/main/java/org/dcm4che3/net/Device.java
```diff
@@ -1084,7 +1063,11 @@ public ApplicationEntity getApplicationEntity(String aet) {
...
+        applicationEntitiesMap.values().forEach(ae -> aeTitles.addAll(ae.getAETitleAliases()));
+        return aeTitles;
```
msiemczyk, line 1070: Nit: maybe add blank line before return to make it a bit more readable?

dcm4che-net/src/main/java/org/dcm4che3/net/Device.java
```diff
@@ -1164,6 +1147,19 @@ public final TrustManager getTrustManager() {
         return tm;
     }
 
+    private ApplicationEntity getApplicationEntityOrByAETitleAlias(String aet) {
+        ApplicationEntity ae = applicationEntitiesMap.get(aet);
+
+        if (ae == null) {
+            //check alias AE titles
+            return applicationEntitiesMap.values().stream()
+                                         .filter(applicationEntity -> applicationEntity.getAETitleAliases().contains(aet))
```
msiemczyk, line 1156: Not sure why IntelliJ keeps doing that, but line continuation should be only double tabbed; shifting code like that just reduces the amount of available space left...remember lines shouldn't be longer than 120 characters. I thought we fixed that?

https://google.github.io/styleguide/javaguide.html

dcm4che-net/src/test/java/org/dcm4che3/net/DeviceTest.java
```diff
@@ -37,50 +37,88 @@
...
+    private ApplicationEntity ae;
+
     @Test
     public void testReconfigure() throws Exception {
         Device d1 = createDevice("test", "AET1");
         Device d2 = createDevice("test", "AET2");
         d2.setOlockHash("I.hash.you.not");
-        d2.setStorageVersion(35);
-        
+                                                                d2.setStorageVersion(35);
```
msiemczyk, line 75: what happened here?

dcm4che-net/src/test/java/org/dcm4che3/net/DeviceTest.java
```diff
@@ -37,50 +37,88 @@
...
 
-    /**
-     * Test method for {@link org.dcm4che3.net.Device#reconfigure(org.dcm4che3.net.Device)}.
-     */
+    private static final String AE_TITLE = "MyAeTitle";
+    private static final String FIRST_ALIAS_AE_TITLE = "AliasAeTitle";
+    private static final String SECOND_ALIAS_AE_TITLE = "AnotherAliasAeTitle";
+
+    @Rule
+    public TemporaryFolder temporaryFolder = new TemporaryFolder();
```
msiemczyk, line 66: I don't see this rule being used?

dcm4che-net/src/test/java/org/dcm4che3/net/DeviceTest.java
```diff
@@ -37,50 +37,88 @@
...
+
+        assertThat("Olock Hash", d1.getOlockHash(), equalTo("I.hash.you.not"));
+        assertThat("Storage version", d1.getStorageVersion(), equalTo(35L));
+    }
+
+    @Test
+    public void getApplicationEntity_ReturnsSameApplicationEntity_GivenAeTitleOrAETitleAliases() {
+
+        Device device = createDevice("AnyDeviceName", AE_TITLE);
+        device.getApplicationEntity(AE_TITLE).
```
msiemczyk, line 91: Nit: oops, I think I made this mistake; dot "." should be on the next line.

dcm4che-net/src/test/java/org/dcm4che3/net/DeviceTest.java
```diff
@@ -37,50 +37,88 @@
...
         ApplicationEntity ae = d1.getApplicationEntity("AET2");
         assertNotNull(ae);
         List<Connection> conns = ae.getConnections();
         assertEquals(1, conns.size());
-        
-       assertThat("Olock Hash", d1.getOlockHash(), equalTo("I.hash.you.not"));
-       assertThat("Storage version", d1.getStorageVersion(), equalTo(35L));
-       
+
+        assertThat("Olock Hash", d1.getOlockHash(), equalTo("I.hash.you.not"));
```
msiemczyk, line 83: Now I see what is going on here...we have our IDEs setup for tab being 3 spaces, but original file was 4 spaces per tab...I'm not sure what is the best approach here., but I see that in master even Device class uses 4 so I guess better to conform to their rule here.

dcm4che-net/src/main/java/org/dcm4che3/net/Device.java
agfa88: should be ok

dcm4che-net/src/test/java/org/dcm4che3/net/DeviceTest.java
agfa88: not sure we have 4 spaces for tab

dcm4che-net/src/main/java/org/dcm4che3/net/Device.java
```diff
@@ -1164,6 +1171,27 @@ public final TrustManager getTrustManager() {
         return tm;
     }
 
+    private final ApplicationEntity findApplicationEntityByAeTitleOrAlias(String aet) {
```
hczedik, line 1174: Not sure if this method could be used by multiple threads concurrently?
Then you would have to think a bit more about correct synchronization.

hczedik (approved):

msiemczyk (approved):

PR #1161, opened 2022-04-07
IEI-187643 The transient field aliasApplicationEntitiesMap in Device …

msiemczyk:
…class is null in the returned device instances from remote EJB

- Changed the ApplicationEntity setAETitleAliases to be more in-line with master code by accepting String varargs argument instead of List and added new unit tests for this functionality.
- Cleaned up some code in Device.
- Reworked the newly added DeviceTest to more definitive and added the missing cases.

### DIFF
--- a/dcm4che-net/src/main/java/org/dcm4che3/net/ApplicationEntity.java
+++ b/dcm4che-net/src/main/java/org/dcm4che3/net/ApplicationEntity.java
@@ -38,6 +38,23 @@
 
 package org.dcm4che3.net;
 
+import java.io.IOException;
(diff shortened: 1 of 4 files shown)


msiemczyk:
Hi @hczedik,

I tried to create PR of my changes to Zhirong's fork so I can add my commit his PR, but I think there is something wrong with his setup or permissions, because I couldn't see his fork. I cleanup some of his code and reworked the unit tests mainly. One main production change is change to setAETitleAliases in ApplicationEntity so it is same as master branch and also it is easier to use.

hczedik (approved):

PR #1163, opened 2022-04-08
IEI-187643 Fixing the configuration bean reflection issue

msiemczyk:
- Fixed the capitalization of setAeTitleAliases and getAeTitleAliases, as well as change setter back to List, in ApplicationEntity so configuration property name can be found in property descriptor and serialized.
- Added isAeTitleAlias to ApplicationEntity in order to use in Device's stream that looks up AEs to avoid penalty of coping the List.


greg-knox (approved):

gunterze:
Are you aware of property `OtherAETitles` introduced by 2921990d58111bda5aa382ddf0b0536605367708 ?

Issue #787, opened 2020-09-09
Attributes.(g|s)etDate[Range] does not consider (0008,0201) Timezone Offset From UTC for DA attributes


gunterze:
Attributes.(g|s)etDate[Range] does not consider
- (0008,0201) Timezone Offset From UTC
- default Timezone specified by `Attributes.setDefaultTimeZone(TimeZone)`

for attributes with _VR_ `DA`

JanHoracek:
Hello, 

wouldn't it cause problems when you try to get patient birth date, study date etc.?
When the images are created during summer time, but they are received by PACS few days later when it's winter time already, the date can be wrong.

Example: 
1. Patient birth date/time or study date/time is "2020-10-25 00:15:00" based on CEST, which is UTC + 2.
2. When you save it to PACS few hours later, when PACS switches to winter time (UTC + 1), getDate(...) returns "2020-10-24" because it will be converted into "2020-10-24 23:15:00"

The same happens when you receive images from different parts of the world. 

I propose that this behaviour should be parametrized and the default behaviour should be the "old" one, i.e. do not take into consideration the timezone...

BTW: There were similar issues already few years ago at DCM4CHE 3: https://dcm4che.atlassian.net/browse/LIB-420 and https://dcm4che.atlassian.net/browse/LIB-312

Kind regards

Jan

gunterze:
The timezone is applied to the stored/accessed value, not to the time when it's stored/accessed.

Sent with [ProtonMail](https://protonmail.com) Secure Email.

‐‐‐‐‐‐‐ Original Message ‐‐‐‐‐‐‐

JanHoracek:
I understand that it's applied to the stored value. For example DICOM contains following attributes:

```
(0008,0201)="+0700"
(0010,0030)="20180501"
(0010,0032)="001526.000"
```

My local timezone is "+0100"

When I call 
```Date birthDate = attrs.getDate(Tag.PatientBirthDate);```
what date will I receive?

gunterze:
Mon Apr 30 18:00:00 GMT+01:00 2018

The problem is rather caused by using Date as return type - when I start working on dcm4che-3/5, there was not yet package java.time available which provides specific types for Temporals with/without date, time and timezone information.

hczedik:
Would actually be really nice to add methods that return the new java.time types eventually. I think it could also still be done with dcm4che3/5 in a backwards compatible way (but just providing additional methods).

gunterze:
[...]

Issue #928, opened 2021-03-22
testSetTimezoneOffsetFromUTC() fails if system timezone has negativ UTC time offset


brentkellohen:
**Describe the bug**
mvn install fails after fresh pull of code and install of Maven 3.6.3 on Manjaro ArchLinux

```
[INFO] Running org.dcm4che3.data.AttributesTest
[ERROR] Tests run: 49, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 0.017 s <<< FAILURE! - in org.dcm4che3.data.AttributesTest
[ERROR] testSetTimezoneOffsetFromUTC(org.dcm4che3.data.AttributesTest)  Time elapsed: 0.005 s  <<< FAILURE!
org.junit.ComparisonFailure: expected:<19[70010]1> but was:<19[69123]1>
        at org.dcm4che3.data.AttributesTest.testSetTimezoneOffsetFromUTC(AttributesTest.java:292)
```

**To Reproduce**
1. Pull dcm4che commit 127af6e
2. Install maven 3.6.3
3. Run mvn install in dcm4che directory

**Expected behavior**
Build successful

**Screenshots**
![image](https://user-images.githubusercontent.com/17730896/112063989-836ae580-8b1f-11eb-91b4-b75b8a844d98.png)

**Desktop (please complete the following information):**
See above

**Additional context**
I did have a version of DCMTK installed and was unsure if this may have conflicted with the installation. This was uninstalled but the error persists. 

Java version
```
openjdk version "15.0.2" 2021-01-19
OpenJDK Runtime Environment (build 15.0.2+7)
OpenJDK 64-Bit Server VM (build 15.0.2+7, mixed mode)
```

gunterze:
What's the system timezone?
```
$ timedatectl
               Local time: Di 2021-03-23 07:55:07 CET
           Universal time: Di 2021-03-23 06:55:07 UTC
                 RTC time: Di 2021-03-23 06:55:07    
                Time zone: Europe/Vienna (CET, +0100)
System clock synchronized: yes                       
              NTP service: active                    
          RTC in local TZ: no   
```

brentkellohen:
![image](https://user-images.githubusercontent.com/17730896/112171078-caed8200-8bb0-11eb-8567-0bba3cfb20e5.png)

gunterze:
Related to #787.

Comment out
```java
    /*
        @Test
        public void testSetTimezoneOffsetFromUTC() throws Exception {

```

PR #1164, opened 2022-04-12
Allow to specify whether to apply timezone offset via DatePrecision

hczedik:
If not specified, by default, it will be applied for VR.DT values and when accessing (get/set) VR.DA and VR.TM values combined, but not when accessing them separately.

Explanation:
The change done in https://github.com/dcm4che/dcm4che/issues/699 is causing quite some problems in our application.
In particular, when accessing a "combined" (long) dateAndTime tag (e.g. Tag.StudyDateAndTime) the current handling is quite inconsistent: When setting a Date, the VR.TM value will be timezone corrected, but the VR.DA value will not (!), which is incorrect if the timezone offset goes over midnight to the next/previous day.
When getting a Date, the offset will be applied for both the VR.DA and VR.TM values, making the behavior even more inconsistent. (Set and get methods should behave consistently.)

With these changes, I want to make this behavior consistent. Please see the new unit tests.
When accessing date and time combined (long tag), then the correction will always happen by default.
When accessing them separately, then no correction will be done by default (because it is impossible to do it correctly).

Additionally, using a new field in DatePrecision, the behavior can also be controlled by the user.

I know that ideally we should eventually introduce methods which work with the new java.time types, as you mentioned in https://github.com/dcm4che/dcm4che/issues/787
That would be nice of course! But until then, I believe that these changes already improve the behavior a lot, and allow the user quite some flexibility.

### DIFF
--- a/dcm4che-core/src/main/java/org/dcm4che3/data/Attributes.java
+++ b/dcm4che-core/src/main/java/org/dcm4che3/data/Attributes.java
@@ -1251,7 +1251,7 @@ public class Attributes implements Serializable {
     }
 
     public Date getDate(int tag) {
-        return getDate(null, tag, null, 0, null, new DatePrecision());
+        return getDate(null, tag, null, 0, null, null);
     }
 
     public Date getDate(int tag, DatePrecision precision) {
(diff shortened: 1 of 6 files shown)


hczedik:
Note that (I think) I am also fixing https://github.com/dcm4che/dcm4che/issues/928 with this. I have commented back in the corresponding unit test.

hczedik:
Hi @gunterze,

did you have some time to take a look at these changes already?
We need a fix here, at least for the inconsistency between the methods getDate and setDate with "combined" (long) tags, as this is causing quite some regression in our application.

Thanks.

gunterze:
Prefer #1172

hczedik:
Thanks @gunterze, your changes in https://github.com/dcm4che/dcm4che/issues/1172 also work for us.

Issue #706, opened 2020-04-16
Modify Attributes.addAll() to preserve multiple Private Creator Data Elements with equal values in one private group


gunterze:
Multiple Private Creator Data Elements with equal values in one private group should not be used accord [DICOM PS 3.5, 7.8.1 Private Data Element Tags](http://dicom.nema.org/medical/dicom/current/output/html/part05.html#sect_7.8.1):

> Private Creator Data Elements numbered (gggg,0010-00FF) (gggg is odd) shall be used to reserve a block of Elements with Group Number gggg for use by an individual implementer. The implementer shall insert an identification code in the first unused (unassigned) Element in this series to reserve a block of Private Elements. The VR of the private identification code shall be LO (Long String) and the VM shall be equal to 1. **A Private Creator identifier may be used only once within a Group; reserving multiple blocks of Elements in the same Group with the same identifier is not allowed**. The Private Creator Data Elements shall only contain characters from the Default Character Repertoire and not an Extended or Replacement Character Repertoire, even though the LO VR is one that is affected by the Specific Character Set (0008,0005).

But there are vendors generating such Datasets:
![Screenshot from 2020-04-16 16-07-31](https://user-images.githubusercontent.com/400095/79467002-961a9780-7ffd-11ea-9368-1c20b591873a.png)

Related #40

PR #1167, opened 2022-04-19
public tags were incorrectly handled after merging privateCreators wi…

Stromausfall:
Related to https://github.com/dcm4che/dcm4che/issues/706, when one Attribute instance with a privateCreator is added to another Attribute instance with a privateCreator - both using the same tag but different values.
Then if a public tag follow these privateCreator tags an exception is thrown:

(0040,0254) is not a private Data Element
java.lang.IllegalArgumentException: (0040,0254) is not a private Data Element
	at org.dcm4che3.data.Attributes.creatorTagOf(Attributes.java:505)
	at org.dcm4che3.data.Attributes.set(Attributes.java:2179)
	at org.dcm4che3.data.Attributes.add(Attributes.java:2471)
	at org.dcm4che3.data.Attributes.addAll(Attributes.java:2228)

This appears to be due to the privateCreator0 variable not being reset to 'null' on public tags.

### DIFF
--- a/dcm4che-core/src/main/java/org/dcm4che3/data/Attributes.java
+++ b/dcm4che-core/src/main/java/org/dcm4che3/data/Attributes.java
@@ -2422,6 +2422,7 @@ public class Attributes implements Serializable {
             } else {
                 creatorTag = 0;
                 privateCreator = null;
+                privateCreator0 = null;
             }
 
             if (selection != null && !selection.contains(privateCreator, tag))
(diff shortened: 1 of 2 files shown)


hczedik:
Hi @gunterze,

@Stromausfall and myself worked together on this fix.
The bug is causing quite some regression in our application.
Can you please take a quick look at this one? It is a small and simple fix.

gunterze (approved):
Thx!

PR #1179, opened 2022-05-09
add an option to create message filename with UUID (instead of msh 9)

ariel-rimon-21:


gunterze:
Just shortened --use-uuid-for-filename to --uuid 05ca041d775372ed4798fcc759c438adf02ad08d

PR #1182, opened 2022-05-10
Update DicomInputStream: add method name to warning when skipping att…

craigtmoore:
…ributes.

 * this makes it easier to figure out which method is generating the message

### DIFF
--- a/dcm4che-core/src/main/java/org/dcm4che3/io/DicomInputStream.java
+++ b/dcm4che-core/src/main/java/org/dcm4che3/io/DicomInputStream.java
@@ -840,8 +840,8 @@ public class DicomInputStream extends FilterInputStream
     private void skipAttribute(String message, String methodName) throws IOException {
         String tagAsString = TagUtils.toString(this.tag);
         LOG.warn(message, tagAsString, length, tagPos, methodName);
-        long skipLength = skip(length);
-        LOG.debug("Skipped {} actual bytes for {}", skipLength, tagAsString);
+        skipFully(length);
+        LOG.debug("Skipped {} bytes for {}", length, tagAsString);
     }
 
     private void readSequence(int len, Attributes attrs, int sqtag)
--- a/dcm4che-core/src/main/java/org/dcm4che3/io/DicomInputStream.java
+++ b/dcm4che-core/src/main/java/org/dcm4che3/io/DicomInputStream.java
@@ -68,14 +68,14 @@ public class DicomInputStream extends FilterInputStream
         LoggerFactory.getLogger(DicomInputStream.class);
 
     private static final String UNEXPECTED_NON_ZERO_ITEM_LENGTH =
-        "Unexpected item value of {} #{} @ {}";
+        "Unexpected item value of {} #{} @ {} during {}";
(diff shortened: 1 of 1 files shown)


dcm4che-core/src/main/java/org/dcm4che3/io/DicomInputStream.java
```diff
@@ -831,10 +832,16 @@ private void checkIsThis(DicomInputStream dis) {
...
+    /**
+     * @param message the message to use in the warning log message
+     * @param methodName the name of the method that is skipping the attribute
+     * @throws IOException potentially thrown when performing the 'skip' operation
+     */
+    private void skipAttribute(String message, String methodName) throws IOException {
+        String tagAsString = TagUtils.toString(this.tag);
+        LOG.warn(message, tagAsString, length, tagPos, methodName);
+        long skipLength = skip(length);
+        LOG.debug("Skipped {} actual bytes for {}", skipLength, tagAsString);
```
hczedik, line 844: I'll out my comment from https://github.com/dcm4che/dcm4che/pull/1152 again here:

Hmm, I wonder if this call to skip(.) here shouldn't actually be a call to skipFully(.)? (@gunterze)

hczedik (approved):

dcm4che-core/src/main/java/org/dcm4che3/io/DicomInputStream.java
gunterze: Yes, replacing skipFully() by skip() actually changes the behavior!

hczedik: Ah, yes indeed! I didn't notice that on master there was already skipFully here, because on the generic-config branch there was also a skip() here before. On the generic-config branch, the fix for this (https://github.com/dcm4che/dcm4che/pull/404) was never applied!

@craigtmoore: instead of now breaking it on master again, I would suggest to actually port the fix back to your generic-config branch.

craigtmoore: OK, I've changed it back to skipFully(). Good catch!

dcm4che-core/src/main/java/org/dcm4che3/io/DicomInputStream.java
```diff
@@ -831,10 +832,16 @@ private void checkIsThis(DicomInputStream dis) {
...
+    /**
+     * @param message the message to use in the warning log message
+     * @param methodName the name of the method that is skipping the attribute
+     * @throws IOException potentially thrown when performing the 'skip' operation
+     */
+    private void skipAttribute(String message, String methodName) throws IOException {
+        String tagAsString = TagUtils.toString(this.tag);
+        LOG.warn(message, tagAsString, length, tagPos, methodName);
         skipFully(length);
+        LOG.debug("Skipped {} bytes for {}", length, tagAsString);
```
hczedik, line 844: Okay, now it is already merged. but as also noted in https://github.com/dcm4che/dcm4che/pull/1184#discussion_r872455148 I think this debug log message does not provide any additioal value, as all the information is anyways already contained in the warn log message above.

PR #1186, opened 2022-05-17
IEI-190922 - Added a new exception type that signifies that the excep…

djbrown67:
Added a new exception type that signifies that the exception doesn't need its stack trace logged by DicomServiceRegistry.

We have some application logic where it's possible that DicomServiceRegistry might get an exception for every image in an association. In our application we log a warning with details about what caused the exception.
However, when the exception arrives in DicomServiceRegistry it gets logged again (not in itself a problem) along with a stack trace. We don't want the stack trace logging since (a) it's not particularly helpful in this case and (b) it very quickly results in massive log files.

In order to not change this for every use case we've introduced a new subclass of the DicomServiceException for cases where applications have already done some kind of handling of the exception and that DicomServiceRegistry can identify and handle differently.


hczedik (approved):

hczedik:
Hi @gunterze,
may we ask you to give this PR here a look?
We have already committed the same change to the generic-config branch (see https://github.com/dcm4che/dcm4che/pull/1196 ).

hczedik:
Remove @greg-knox as reviewer, because he has already approved the corresponding change for generic-config (https://github.com/dcm4che/dcm4che/pull/1196).

@gunterze, can you merge this please?

PR #1194, opened 2022-05-30
IEI-196571 decrement the performing operations counter when the pduencoder throws an exception

mdoris:
This will allow the Association to be closed and the thread it is consuming to be returned to the thread pool.


hczedik:
> i believe Hermann is on paternity leave

I will be from July to October (2022). I've asked Lukas to create a github account and act as my replacement for such reviews.

PR #1197, opened 2022-06-03
Prevent collision when renaming same file concurrently fix #1189

### LINKED ISSUE #1189: File access collision when receiving same dicom file simultaneously in StoreSCP
**Describe the bug**

When the same dicom file is received in parallel, the multiple threads handling it try to access and write the same destination file simultaneously, hence error happen.

**To Reproduce**
This simple class outputs many errors in the logs

```
 import java.io.File;

 import org.dcm4che3.tool.storescp.StoreSCP;
 import org.dcm4che3.tool.storescu.StoreSCU;

 import lombok.extern.slf4j.Slf4j;

 /** This class is for testing *manually* the SCP Server */
 public class ManualTester2 {

	public static void main(String[] args) throws Exception {

		int serverPort = 11_112;
		String[] scpArgs = new String[] { "-b",
				"aaa@127.0.0.1:11112",
				"--accept-unknown" };
		StoreSCP.main(scpArgs);

		for (int i = 0; i < 10; i++) {
			Thread t = new Thread(() -> {
				File dcmFilesToSendDir = new File("/path/to/your/single/file.dcm");
				String[] scuArgs = new String[] { "-c", "aaa@127.0.0.1:" + serverPort, dcmFilesToSendDir.toString() };
				StoreSCU.main(scuArgs);
			});
			t.start();
		}

		Thread.sleep(10000);
	}
}
```

**Expected behavior**
No errors should be shown, each file sent should overwrite the previous one silently if they are the same.

**Desktop (please complete the following information):**
The behavior was observed on Windows and Linux with dcm4che 5.25.2 and 5.24.2

**Additional context**
I'm not sure how often simultaneously receiving the same file happens in the real world but I face this problem at a partner's site, hence, I'm inclined to think it happens sometimes.

Additionally, I feel like the previous attempts to resolve this issue ( https://github.com/dcm4che/dcm4che/issues/1112 ) was not entirely sufficient: `File.createTempFile(iuid, PART_EXT, storageDir);` was enough to prevent collisions on the temporary file, but collisions still happen on renaming the temp file to the final file

**Workaround**

The workaround I propose (and will propose as a PR) is to just add the `synchronized` keyword to the "renameTo" method.
Hence

``` private static void renameTo(Association as, File from, File dest)```

would become

``` private synchronized static void renameTo(Association as, File from, File dest)```

I tested this fix and it does the job.
An alternative (maybe faster) would be to keep in a List the last 100 or so filenames and lock only on the filenames so that we would do lock for every file but only for real collisions, but I feel like it would be a lot more complex both to explain and maintain. And performance gain would have to be proven.

fmarot:
this fix prevents the renameTo() method to be accessed concurrently, because it generated errors in logs and potentially file corruption when accessing the same file.

### DIFF
--- a/dcm4che-tool/dcm4che-tool-storescp/src/main/java/org/dcm4che3/tool/storescp/StoreSCP.java
+++ b/dcm4che-tool/dcm4che-tool-storescp/src/main/java/org/dcm4che3/tool/storescp/StoreSCP.java
@@ -156,7 +156,8 @@ public class StoreSCP {
         }
     }
 
-    private static void renameTo(Association as, File from, File dest)
+    /** method is synchronized to prevent (rare) concurrent access to the same file */
+    private synchronized static void renameTo(Association as, File from, File dest)
             throws IOException {
         LOG.info("{}: M-RENAME {} to {}", as, from, dest);
         if (!dest.getParentFile().mkdirs())


gunterze:
Solved by https://github.com/dcm4che/dcm4che/commit/2100085fea8f1cda915761b752b2183075b3b7f6 committed by @or150

PR #1202, opened 2022-06-13
Decrement performing ops on exception while writing response Fix #1201

### LINKED ISSUE #1201: Association performing operations counter does not decrement when exception is encountered when writing response
Association class performing counter track the number of outstanding requests against an Association. When the association is being torn down, there is a wait for outstanding operations to complete / for the performing counter to reach zero.

When the response is attempted to be written but an exception occurs, the exception will prevent the performing counter from being decremented resulting in stuck threads.

**To Reproduce**
1. Open an association and send in a request 
2. Introduce an exception (socket or otherwise) while the response is being written to the socket

**Expected behavior**
For the performing operations counter to decrement allowing the Association to be torn down and free resources.

mdoris:
Ensures the decrement of the counter always gets performed even when the response cannot be written.

Also adjusts the log level when logging an exception while attempting to process a request in order to retain in error logs to allow for logs to retrieved before the exception stacktrace rolls out of the server logs.

### DIFF
--- a/dcm4che-net/src/main/java/org/dcm4che3/net/service/DicomServiceRegistry.java
+++ b/dcm4che-net/src/main/java/org/dcm4che3/net/service/DicomServiceRegistry.java
@@ -58,7 +58,7 @@ import org.dcm4che3.net.pdu.PresentationContext;
 public class DicomServiceRegistry implements DimseRQHandler {
 
     private final HashMap<String, DimseRQHandler> services =
-            new HashMap<String, DimseRQHandler>();
+            new HashMap<>();
 
     public void addDicomService(DicomService service) {
         addDimseRQHandler(service, service.getSOPClasses());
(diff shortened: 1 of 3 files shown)


hczedik:
This fixes: https://github.com/dcm4che/dcm4che/issues/1201 (not sure why github did not link it automatically so far).

dcm4che-net/src/main/java/org/dcm4che3/net/service/DicomServiceRegistry.java
```diff
@@ -85,7 +85,7 @@ public void onDimseRQ(Association as, PresentationContext pc,
         try {
             lookupService(as, dimse, cmd).onDimseRQ(as, pc, dimse, cmd, data);
         } catch (DicomServiceException e) {
-            Association.LOG.info("{}: processing {} failed. Caused by:\t",
+            Association.LOG.error("{}: processing {} failed. Caused by:\t",
```
gunterze, line 88: The Exception may be caused by a bad request => prefer to stay with INFO level.

mdoris: While it may be caused by a bad request, often we are asked to look at why we reject/return an error as a start point to finding the bad request. In some sites the info/warn level logs are rolling out of the logs before we are able to collect the logs resulting in delays in diagnosing the problems while at error level the logs will be retained longer. Additionally, error level is consistent with a non-recoverable error condition vs an informational message.

However, I will return the log level to info should you feel strongly about it.

mdoris: That argument would lend toward the need to add auditing as it is a non-recoverable processing error. Changing the log level doesn't stop that and the lower level does make the diagnosis of the problem more difficult. 

However, I have reverted the log level as per your request.

PR #1204, opened 2022-06-16
Decrement performing ops when exception occurs writing response

mdoris:
When writing the dimse response but an exception occurs, decrement the performing operations counter so that the Association can be torn down. Also, changes the log level when an exception occurs while performing a dimse request so that the errors appear in the error log and not roll out of the logs too quickly.

### DIFF
--- a/dcm4che-net/src/test/java/org/dcm4che3/net/AssociationTest.java
+++ b/dcm4che-net/src/test/java/org/dcm4che3/net/AssociationTest.java
@@ -7,6 +7,7 @@ import java.net.Socket;
 import java.net.SocketException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
+import java.util.concurrent.Semaphore;
 
 import org.dcm4che3.data.Attributes;
 import org.dcm4che3.data.Tag;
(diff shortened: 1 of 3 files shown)


alan-farquhar (approved):

dcm4che-net/src/test/java/org/dcm4che3/net/AssociationTest.java
```diff
@@ -0,0 +1,106 @@
...
+        }
+    }
+
+    private class DummyInputStream extends InputStream {
+
+        Semaphore semaphore = new Semaphore(-1);
+
+        @Override public int read() throws IOException {
+            try {
+                // Just a short sleep to simulate the blocking read for nextPDU
```
hczedik, line 89: Your comment says "a shot sleep", but in fact this semaphore will block infinitely, doesn't it?

mdoris: Originally it was a sleep that was changed to a semaphore based on the sonarqube result. Both are true, the semaphore does block indefinitely, but it does end up just being for a moment.

mdoris: Hm... you're right. I should be using shutdownNow() instead of shutdown() on the executor service. I will correct and also update that comment.

PR #1207, opened 2022-06-21
Adding XRay3DAngiographic to the MultiframeExtractor storage classes

TimMullender:
We'd like to be able to extract multiframe DICOMs that are using the X-Ray 3D Angiographic Storage Class (1.2.840.10008.5.1.4.1.1.13.1.1) with dcm4che. Our initial testing internally suggests this change should make that possible.

There don't appear to be any tests around MultiframeExtractor that I could find, is there somewhere I should be adding one?


gunterze (approved):
Thx!

PR #1209, opened 2022-06-22
Update audit logger reconfigure to include doIncludeInstanceUID

marryatma:


msiemczyk:
I guess you could add unit test for this class if you really wanted to ;)

msiemczyk (approved):

hczedik (approved):

hczedik:
Would be good to look into why there are differences between AuditLogger on the master branch and our generic-config branch, and to see whether we can align the two.

dcm4che-net-audit/pom.xml
```diff
@@ -40,5 +40,15 @@
       <type>test-jar</type>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>junit</groupId>
+      <artifactId>junit</artifactId>
+    </dependency>
+    <dependency>
+       <groupId>org.hamcrest</groupId>
+       <artifactId>hamcrest-library</artifactId>
```
msiemczyk, line 49: Hamcrest usually goes before junit because junit library has subset of hamcrest matchers and they collide with each other in Eclipse IDE. Also typically we use "hamrecrest-all" jar.

dcm4che-net-audit/src/test/java/org/dcm4che3/net/audit/AuditLoggerTest.java
```diff
@@ -0,0 +1,356 @@
...
+ * @author Marryat Ma
+ */
+public class AuditLoggerTest {
+
+    private AuditLogger newLogger;
+    private AuditLogger logger;
+
+    @Before
+    public void before() {
+        newLogger = new AuditLogger();
```
msiemczyk, line 62: This and next new can be moved to member declaration (so fields are final, unless there is reason not to). Test classes are created for each test so it makes no difference to have it here or in before().

dcm4che-net-audit/src/test/java/org/dcm4che3/net/audit/AuditLoggerTest.java
```diff
@@ -0,0 +1,356 @@
...
+
+        logger = new AuditLogger();
+        logger.setDevice(new Device());
```
msiemczyk, line 65: Unless there is a good reason to use real Device I would prefer that use a (power) mock of unit testing as it will avoid testing "device.reconfigureConnections" in these tests.

dcm4che-net-audit/src/test/java/org/dcm4che3/net/audit/AuditLoggerTest.java
```diff
@@ -0,0 +1,356 @@
...
+    }
+
+    @Test
+    public void reconfigure_UpdatesFacility_WhenCalled() {
```
msiemczyk, line 69: I think test names would benefit from changing "WhenCalled" to specific condition. I know some people say that you should only have one assert per tests, but in this situation I don't this applies. If you are just setting one member other should be null so don't you want to assert That other fields are still null?

I think better approach here would be to all of the fields and assert on all of them. And only have additional tests for branches (if in clearActiveConnectionsAndClose), which I wouldn't really ask for to be honest.

dcm4che-net-audit/src/test/java/org/dcm4che3/net/audit/AuditLoggerTest.java
```diff
@@ -0,0 +1,356 @@
...
+
+        assertThat("auditLoggerInstalled is not as expected", logger.getAuditLoggerInstalled(), is(false));
+    }
+
+    @Test
+    public void reconfigure_UpdatesAuditRecordRepositoryDevices_WhenCalled() {
+        final Connection auditUDP = new Connection("audit-udp", "localhost", 514);
+        auditUDP.setProtocol(Connection.Protocol.SYSLOG_UDP);
+
+        final AuditRecordRepository arr1 = new AuditRecordRepository();
```
msiemczyk, line 284: Nit: using numbered suffixes is not a good practice. You could consider more meaningful variable names.

dcm4che-net-audit/src/test/java/org/dcm4che3/net/audit/AuditLoggerTest.java
```diff
@@ -0,0 +1,356 @@
...
+        arr1.addConnection(auditUDP);
+
+        final Device arrDevice1 = new Device("syslog");
+        arrDevice1.addDeviceExtension(arr1);
+        arrDevice1.addConnection(auditUDP);
+
+        newLogger.setAuditRecordRepositoryDevices(Collections.singletonList(arrDevice1));
+
+        final Connection auditTCP = new Connection("audit-tcp", "localhost", 516);
```
msiemczyk, line 293: You could consider creating private helper method to create connections to reduce arrange code clutter (since it also does not matter for this tests what is actually in connections).

dcm4che-net-audit/src/test/java/org/dcm4che3/net/audit/AuditLoggerTest.java
```diff
@@ -0,0 +1,356 @@
...
+
+        final Device arrDevice2 = new Device("syslog2");
+        arrDevice2.addDeviceExtension(arr2);
+        arrDevice2.addConnection(auditTCP);
+
+        logger.setAuditRecordRepositoryDevices(Collections.singletonList(arrDevice2));
+
+        logger.reconfigure(newLogger);
+
+        final Device toDevice = logger.getAuditRecordRepositoryDevices().get(0);
```
msiemczyk, line 307: Why not just use getAuditRecordRepositoryDevices() and compare List directly? Matchers can do equalTo on lists with reference check and since setAuditRecordRepositoryDevices assigns same List is same object instance so that would be better check than just getting first item.

dcm4che-net-audit/src/test/java/org/dcm4che3/net/audit/AuditLoggerTest.java
```diff
@@ -0,0 +1,356 @@
...
+        assertThat("auditRecordRepositoryDevices' device name is not as expected", toDevice.getDeviceName(), is(arrDevice1.getDeviceName()));
+        assertThat("auditRecordRepositoryDevices' device extension is not as expected", toDevice.getDeviceExtension(AuditRecordRepository.class), samePropertyValuesAs(arr1));
+        assertThat("auditRecordRepositoryDevices' connections is not as expected", toDevice.getConnections().get(0), samePropertyValuesAs(auditUDP));
```
msiemczyk, line 310: This is really testing Device so I don't think you need to do this.

dcm4che-net-audit/src/test/java/org/dcm4che3/net/audit/AuditLoggerTest.java
```diff
@@ -0,0 +1,356 @@
...
+        newCriteria.setCommonName("NewCommonName");
+        newLogger.setAuditSuppressCriteriaList(Collections.singletonList(newCriteria));
+
+        final AuditSuppressCriteria oldCriteria = new AuditSuppressCriteria();
+        oldCriteria.setCommonName("OldCommentName");
+        logger.setAuditSuppressCriteriaList(Collections.singletonList(oldCriteria));
+
+        logger.reconfigure(newLogger);
+
+        assertThat("auditSuppressCriteriaList is not as expected", logger.getAuditSuppressCriteriaList().get(0), samePropertyValuesAs(newCriteria));
```
msiemczyk, line 325: Not sure why you are using here this matcher.

" Creates a matcher that matches when the examined object has values for all of its JavaBean properties that are equal to the corresponding values of the specified bean."

http://hamcrest.org/JavaHamcrest/javadoc/1.3/org/hamcrest/beans/SamePropertyValuesAs.html

This will compare fields, which is unrelated to the AuditLogger class, which simply clears the list and adds new filters. You should be able to use equalTo() and compare the list, which will iterate through list and do a object reference comparison.

dcm4che-net-audit/src/test/java/org/dcm4che3/net/audit/AuditLoggerTest.java
```diff
@@ -0,0 +1,356 @@
...
+        assertThat("auditRecordRepositoryDevices' device name is not as expected", toDevice.getDeviceName(), is(arrDevice1.getDeviceName()));
+        assertThat("auditRecordRepositoryDevices' device extension is not as expected", toDevice.getDeviceExtension(AuditRecordRepository.class), samePropertyValuesAs(arr1));
+        assertThat("auditRecordRepositoryDevices' connections is not as expected", toDevice.getConnections().get(0), samePropertyValuesAs(auditUDP));
+    }
+
+    @Test
+    public void reconfigure_UpdatesAuditSuppressCriteriaList_WhenCalled() {
+        final AuditSuppressCriteria newCriteria = new AuditSuppressCriteria();
+        newCriteria.setCommonName("NewCommonName");
+        newLogger.setAuditSuppressCriteriaList(Collections.singletonList(newCriteria));
```
msiemczyk, line 317: Is there a benefit to using singletonList vs JDK Arrays.asList()?

dcm4che-net-audit/src/test/java/org/dcm4che3/net/audit/AuditLoggerTest.java
```diff
@@ -0,0 +1,356 @@
...
+    }
+
+    @Test
+    public void reconfigure_UpdatesDoIncludeInstanceUID_WhenCalled() {
+        newLogger.setDoIncludeInstanceUID(true);
+        logger.setDoIncludeInstanceUID(false);
+
+        logger.reconfigure(newLogger);
+
+        assertThat("doIncludeInstanceUID is not as expected", logger.getDoIncludeInstanceUID(), is(true));
```
msiemczyk, line 335: In my opinion, long comments on asserts typically don't make it more readable. The hamcrest will say "Expected: blah, but got <blah>." I would usually just put a field name or thing you are asserting on the message and that is sufficient in my opinion.

dcm4che-net-audit/src/test/java/org/dcm4che3/net/audit/AuditLoggerTest.java
```diff
@@ -0,0 +1,356 @@
...
+    }
+
+    @Test
+    public void reconfigure_UpdatesConnections_WhenCalled() {
```
msiemczyk, line 339: This would be a lot better tested with mock where you simply verify that device.reconfigureConnections(connections, from.connections) was made. So you are not testing Device, that should be done in DeviceTest really.

dcm4che-net-audit/src/test/java/org/dcm4che3/net/audit/ConfigurablePropertiesTest.java
```diff
@@ -55,7 +55,10 @@ protected Class[] getClassesToTest() {
     }
 
     @Override
-    protected Set<Field> getPropertiesToSkip() {
-        return new HashSet<>();
+    protected Set<Field> getPropertiesToSkip() throws NoSuchFieldException {
```
msiemczyk, line 58: Not sure why this is showing up in your PR as you didn't touch those classes? Also my class looks different, I'm confused.

https://github.com/msiemczyk/dcm4che/blob/generic-config/dcm4che-net/src/test/java/org/dcm4che3/net/ConfigurablePropertiesTest.java

dcm4che-net-audit/src/test/java/org/dcm4che3/net/audit/AuditLoggerTest.java
marryatma: If you only have one item in the list, SonarLint prefers using singletonList, not Arrays.asList().

dcm4che-net-audit/src/test/java/org/dcm4che3/net/audit/ConfigurablePropertiesTest.java
marryatma: I needed to add AuditLoggerTest.java in the test directory but noticed that there was an existing ConfigurablePropertiesTest  file that wasn't in the correct test directory and hence was never run.  When I moved ConfigurablePropertiesTest file into the proper directory, the test failed and hence the changes.

msiemczyk: Ah, I didn't see that the file was move. Maybe it would make more sense to combine both files then? i.e. just move the classes and Fields to skip into "net/ConfigurablePropertiesTest.java"?

dcm4che-net-audit/src/test/java/org/dcm4che3/net/audit/AuditLoggerTest.java
marryatma: powermock isn't used in generic-config, so I'll use easymock instead.

msiemczyk: You might get away with that, depending if methods are final or not. I've ran into this with Device (and Connection) class where method is final EasyMock won't be able to mock it. Powermock can do it (or mockito 3.x or maybe 2.x). If you need powermock you just add a dependency to pom.xml.

dcm4che-net-audit/src/test/java/org/dcm4che3/net/audit/AuditLoggerTest.java
marryatma: I'll change to compare the AuditSuppressCriteria.

greg-knox (approved):

PR #1213, opened 2022-06-27
Removed direct access to variables from equals method in Issuer and C…

jay-sukhadiya:
Removed direct access to variables from equals method in Issuer and Code class.
No unit tests were added as it requires Hibernate support (Hibernate session to test proxy). 
We will be adding unit tests in our code as below:
Issuer issuer = createNewIssuer("code");
getHibernateSession().flush();
getHibernateSession().clear();

long issuer_pk = issuer.getPk();
getHibernateSession().clear();
Issuer issuerProxy = getHibernateSession().load(Issuer.class, issuer_pk);

System.out.println(issuerProxy == issuerProxy);
System.out.println(issuerProxy.equals(issuerProxy));

This will return:
true
false

### DIFF
--- a/dcm4che-core/src/main/java/org/dcm4che3/data/Issuer.java
+++ b/dcm4che-core/src/main/java/org/dcm4che3/data/Issuer.java
@@ -39,6 +39,7 @@
 package org.dcm4che3.data;
 
 import java.io.Serializable;
+import java.util.Objects;
 
 import org.dcm4che3.util.StringUtils;
 
@@ -54,7 +55,7 @@ public class Issuer implements Serializable {
     private String universalEntityIDType;
 
     public Issuer(String localNamespaceEntityID, String universalEntityID,
-            String universalEntityIDType) {
+                  String universalEntityIDType) {
         this.localNamespaceEntityID = localNamespaceEntityID;
         this.universalEntityID = universalEntityID;
         this.universalEntityIDType = universalEntityIDType;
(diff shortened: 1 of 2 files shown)


dcm4che-core/src/main/java/org/dcm4che3/data/Issuer.java
```diff
@@ -54,7 +55,7 @@ public class Issuer implements Serializable {
     private String universalEntityIDType;
 
     public Issuer(String localNamespaceEntityID, String universalEntityID,
-            String universalEntityIDType) {
+                  String universalEntityIDType) {
```
alan-farquhar, line 58: @jay-sukhadiya  please undo the tabbing changes here

dcm4che-core/src/main/java/org/dcm4che3/data/Issuer.java
```diff
@@ -187,13 +188,13 @@ public boolean equals(Object o) {
         if (!(o instanceof Issuer))
             return false;
         Issuer other = (Issuer) o;
-        return equals(localNamespaceEntityID, other.localNamespaceEntityID)
-                && equals(universalEntityID, other.universalEntityID)
-                && equals(universalEntityIDType, other.universalEntityIDType);
+        return equals(localNamespaceEntityID, other.getLocalNamespaceEntityID())
```
alan-farquhar, line 191: please undo the getter and setter changes.

PR #1221, opened 2022-07-05
IEI-122058 - Improve audit / syslog thread naming and handling

zliang88:
IEI-122058 - Improve audit / syslog thread naming and handling

Passed NamedThredFactory to ThreadPoolExecutor in SyslogProtocolHandler and added unit test

### DIFF
--- a/dcm4che-net-audit/pom.xml
+++ b/dcm4che-net-audit/pom.xml
@@ -8,6 +8,9 @@
   <artifactId>dcm4che-net-audit</artifactId>
   <name>dcm4che-net-audit</name>
   <packaging>bundle</packaging>
+  <properties>
+    <powermock.version>2.0.9</powermock.version>
+  </properties>
   <build>
     <plugins>
       <plugin>
@@ -43,11 +46,27 @@
     <dependency>
       <groupId>junit</groupId>
       <artifactId>junit</artifactId>
+      <scope>test</scope>
     </dependency>
     <dependency>
       <groupId>org.easymock</groupId>
       <artifactId>easymock</artifactId>
       <version>4.3</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.slf4j</groupId>
+      <artifactId>slf4j-api</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.slf4j</groupId>
+      <artifactId>slf4j-log4j12</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.powermock</groupId>
+      <artifactId>powermock-module-junit4</artifactId>
+      <version>${powermock.version}</version>
       <scope>test</scope>
     </dependency>
   </dependencies>
--- a/dcm4che-net-audit/pom.xml
+++ b/dcm4che-net-audit/pom.xml
@@ -8,9 +8,6 @@
   <artifactId>dcm4che-net-audit</artifactId>
   <name>dcm4che-net-audit</name>
   <packaging>bundle</packaging>
-  <properties>
-    <powermock.version>2.0.9</powermock.version>
-  </properties>
   <build>
     <plugins>
       <plugin>
@@ -56,17 +53,13 @@
     </dependency>
     <dependency>
       <groupId>org.slf4j</groupId>
-      <artifactId>slf4j-api</artifactId>
-    </dependency>
-    <dependency>
-      <groupId>org.slf4j</groupId>
       <artifactId>slf4j-log4j12</artifactId>
       <scope>test</scope>
     </dependency>
     <dependency>
       <groupId>org.powermock</groupId>
       <artifactId>powermock-module-junit4</artifactId>
-      <version>${powermock.version}</version>
+      <version>2.0.9</version>
       <scope>test</scope>
     </dependency>
   </dependencies>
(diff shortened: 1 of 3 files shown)


dcm4che-net-audit/pom.xml
```diff
@@ -8,6 +8,9 @@
   <artifactId>dcm4che-net-audit</artifactId>
   <name>dcm4che-net-audit</name>
   <packaging>bundle</packaging>
+  <properties>
+    <powermock.version>2.0.9</powermock.version>
```
msiemczyk, line 12: I don't think adding a property makes much sense if this is not a group of projects and it is only used once?

dcm4che-net-audit/pom.xml
```diff
@@ -43,12 +46,28 @@
...
       <version>4.3</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.slf4j</groupId>
+      <artifactId>slf4j-api</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.slf4j</groupId>
+      <artifactId>slf4j-log4j12</artifactId>
```
msiemczyk, line 63: Why do you need to add LOG4J binding for tests?

dcm4che-net-audit/src/test/java/org/dcm4che3/net/audit/SyslogProtocolHandlerTest.java
```diff
@@ -0,0 +1,80 @@
...
+ */
+public class SyslogProtocolHandlerTest {
+
+    private static final String AUDIT_SYSLOG_THREAD_NAME_PREFIX = "Audit-Syslog-";
+    private static final int MINIMUM_NUMBER_OF_THREADS = 4;
+
+    @Test
+    public void executor_ThreadNamePrefixesWithCustomizedName_WhenNewThreadIsCreated() {
+
+        ThreadPoolExecutor executor = Whitebox.getInternalState(SyslogProtocolHandler.INSTANCE, "executor");
```
msiemczyk, line 60: Wouldn't it be nicer to use Mockito Spy or even making member protected and making private testable class that exposes the executor with public method? Typically unit testing would somehow test it via public methods. Is there another way to test it; maybe via calling execute?

dcm4che-net-audit/src/test/java/org/dcm4che3/net/audit/SyslogProtocolHandlerTest.java
```diff
@@ -0,0 +1,80 @@
...
+
+        assertThat("Thread name should prefix with Audit-Syslog-",
+                   executor.getThreadFactory()
+                           .newThread(() -> {})
+                           .getName(),
+                   startsWith(AUDIT_SYSLOG_THREAD_NAME_PREFIX));
+    }
+
+    @Test
+    public void executor_ThreadPoolCoreSizeMatchesAvailableProcessorsOrMinimum4_ForSyslogProtocolHandlerThreadPool() {
```
msiemczyk, line 70: This is really the same test as above; i.e. whether constructor sets up members (executor) correctly with the right name and thread pool core size. I personally think it would make more sense to have both asserts in one tests as there are no different conditions here.

dcm4che-net-audit/src/test/java/org/dcm4che3/net/audit/SyslogProtocolHandlerTest.java
```diff
@@ -0,0 +1,80 @@
...
+/**
+ * @author Zhirong Liang <zhirong.liang@agfa.com>
+ */
+public class SyslogProtocolHandlerTest {
+
+    private static final String AUDIT_SYSLOG_THREAD_NAME_PREFIX = "Audit-Syslog-";
+    private static final int MINIMUM_NUMBER_OF_THREADS = 4;
+
+    @Test
+    public void executor_ThreadNamePrefixesWithCustomizedName_WhenNewThreadIsCreated() {
```
msiemczyk, line 58: The "executor_" should be really "constructor" as you are testing constructor. There is no executor method.

dcm4che-net-audit/pom.xml
zliang88: Avoid warning.

dcm4che-net-audit/pom.xml
zliang88: Not big deal. Initially I tried to add more unit tests and had more depedencies. If I have chance to update, either way will work.

dcm4che-net-audit/src/test/java/org/dcm4che3/net/audit/SyslogProtocolHandlerTest.java
zliang88: OK

dcm4che-net-audit/pom.xml
msiemczyk: What warning is that?

zliang88: SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.

msiemczyk: That is not an issue for tests, but it's fine either way.

dcm4che-net-audit/src/test/java/org/dcm4che3/net/audit/SyslogProtocolHandlerTest.java
zliang88: Good question. I don't think it's good idea to change the originl class. Maybe we don't have to unit test  ThreadPoolExecutor. But in order to address the issues that this artifact concerns. I added these 2 tests as I commented in another reply:

Todo :
1) use a named thread pool for auditing

2) verify the number of threads we configure

marryatma (approved):

alan-farquhar (approved):
usually go through master and cherry pick back to generic config.

greg-knox (approved):

lumro (approved):

PR #1234, opened 2022-07-19
Removed the direct access to the variables from the equals method

jay-sukhadiya:
[Class has direct access to variables.docx](https://github.com/dcm4che/dcm4che/files/9141985/Class.has.direct.access.to.variables.docx)

Removed Direct access to variables in the classes that were causing problems while comparing hibernate proxy with itself.
Please refer the attached document.


alan-farquhar (commented):
Approved

PR #1235, opened 2022-07-21
IEI-122058 - Improve audit / syslog thread naming and handling

zliang88:
Added NamedThreadFactory and thread pool to audit syslog

### DIFF
--- a/dcm4che-net-audit/src/main/java/org/dcm4che3/net/audit/SyslogProtocolHandler.java
+++ b/dcm4che-net-audit/src/main/java/org/dcm4che3/net/audit/SyslogProtocolHandler.java
@@ -45,6 +45,10 @@ import java.net.DatagramPacket;
 import java.net.InetAddress;
 import java.net.Socket;
 import java.net.SocketTimeoutException;
+import java.util.concurrent.Executor;
+import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.ThreadPoolExecutor;
+import java.util.concurrent.TimeUnit;
 
 import org.dcm4che3.net.Connection;
 import org.dcm4che3.net.TCPProtocolHandler;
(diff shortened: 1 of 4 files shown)


dcm4che-net-audit/src/main/java/org/dcm4che3/net/audit/SyslogProtocolHandler.java
```diff
@@ -212,8 +246,12 @@ public void run() {
                                 s);
                         break;
                     }
-                    LOG.info("Received Syslog message of {} bytes from {}",
```
alan-farquhar, line 215: Please restore the log.info messages we shouldn't be changing these to debug in dcm4che master.

alan-farquhar (changes requested):
undo log.debug to restore these back to log.info

alan-farquhar (approved):

gunterze (changes requested):
Rejected, because it introduces an additional, not customizable Thread-pool.
Suggest to just provide a static setter to overwrite the use of device.executor by another executor of your choice:
```java
    private static volatile Executor executor;

    public static void setExecutor(Executor executor) {
        SyslogProtocolHandler.executor = executor;
    }

    @Override
    public void onAccept(Connection conn, Socket s) {
        if (executor != null)
            executor.execute(new SyslogReceiverTLS(conn, s));
        else
            conn.getDevice().execute(new SyslogReceiverTLS(conn, s));
    }

    @Override
    public void onReceive(Connection conn, DatagramPacket packet) {
        if (executor != null)
            executor.execute(new SyslogReceiverUDP(conn, packet));
        else
            conn.getDevice().execute(new SyslogReceiverUDP(conn, packet));
    }
```

gunterze (approved):

PR #1252, opened 2022-10-20
IEI-197278 Added new SYSLOG_NONENCRYPTED protcol to Connection to mor…

stephenfrederick:
…e readily support TCP connections that do not use SSL


gunterze:
Why? SYSLOG_TLS actually only use TLS, if you configured at least one cipher - otherwise it's already just plain TCP

stephenfrederick:
> Why? SYSLOG_TLS actually only use TLS, if you configured at least one cipher - otherwise it's already just plain TCP

We need this new protocol type when doing external auditing. Our Web Server needs to know that the protocol is not using TLS by some other method than checking the number of ciphers. By having a separate protocol type this will work without significant rework of our application.

stephenfrederick:
I decided to go in a different direction in the EI code, so this change is no longer required.

Issue #1243, opened 2022-09-09
jpg2dcm error ,When the video exceeds 2G


singworld:
**Describe the bug**
When the video exceeds 2G
jpg2dcm: null

**To Reproduce**
![image](https://user-images.githubusercontent.com/37404619/189290244-868ae763-9776-48aa-aeb6-f81bea27cc31.png)

![image](https://user-images.githubusercontent.com/37404619/189289244-00719cb1-643f-4745-8daf-4c3985382604.png)

![image](https://user-images.githubusercontent.com/37404619/189290655-42879a69-c32e-4d8a-8296-539d526ca744.png)

PR #1266, opened 2022-11-23
Allow length > 2Gib for BulkData

hczedik:
Our service organization is getting more and more cases, where customers have DICOM data with values above the 2Gib limit that dcm4che currently supports.

In one particular case we received a multiframe which is uncompressed and the PixelData is slightly above 2Gib. While it is obviously not very smart to send in such data uncompressed, as it could be represented compressed without problems (and much more efficiently), we cannot really do anything about it as the data comes from a thirdparty and the customer expects us to be able to archive such data.

This PR is now a first suggestion (work in progress!) to fix the most important use cases for us. We want to be able to compress such data, and additionally I changed the SAXWriter a bit, so that it can be successfully represented as XML (with BulkData references).

The compression using the Transcoder seems to work perfectly with those changes (tested with dcm2dcm command).

Note that with the current change, the BulkData reference still has a negative number as length in the URI (as can be tested with the dcm2xml command). This should be probably still be changed.
Also, within the DicomInputStream there are many more places where currently the int length is used, where instead the longLength() should be used.

@gunterze, I would like to get your general feedback on this improvement, before continuing any work on it.
Thanks!

Relates to: https://github.com/dcm4che/dcm4che/issues/1243#issuecomment-1297580711 and IEI-175871

### DIFF
--- a/dcm4che-core/src/main/java/org/dcm4che3/io/DicomInputStream.java
+++ b/dcm4che-core/src/main/java/org/dcm4che3/io/DicomInputStream.java
@@ -535,9 +535,6 @@ public class DicomInputStream extends FilterInputStream
             }
         }
         length = ByteUtils.bytesToInt(buf, 4, bigEndian);
-        if (length < -1) {
-            throw tagValueTooLargeException();
-        }
     }
 
     public boolean readItemHeader() throws IOException {
@@ -655,7 +652,7 @@ public class DicomInputStream extends FilterInputStream
     public void readAttributes(Attributes attrs, int len, Predicate<DicomInputStream> stopPredicate)
             throws IOException {
         boolean undeflen = len == UNDEFINED_LENGTH;
-        long endPos =  pos + (len & 0xffffffffL);
+        long endPos =  pos + toLongLength(len);
         while (undeflen || this.pos < endPos) {
             try {
                 readHeader(stopPredicate);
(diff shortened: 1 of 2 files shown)


SamuelImriska:
I think we should change the type of length to long. I understand  that this is a breaking change, but allowing more then 2GB of data, will require to adjust any code, which builds on top the library anyhow.

hczedik:
Yes, I think that makes sense, @SamuelImriska:

Change the field "length" in DicomInputStream from int to long.

Maybe keep the existing method length() as an int, but mark it as deprecated, and suggest using the new longLength() method.

Same would be nice for BulkData, but that is a bit more tricky, as it is a Serializable so that could break serialized objects if not done carefully.

dcm4che-core/src/main/java/org/dcm4che3/io/DicomInputStream.java
```diff
@@ -678,12 +675,23 @@ public void readAttributes(Attributes attrs, int len, Predicate<DicomInputStream
...
+                    throw tagValueTooLargeException();
+                }
                 handler.readValue(this, attrs);
             } else
                 skipAttribute(UNEXPECTED_ATTRIBUTE, "readAttributes()");
         }
     }
 
+    private static long toLongLength(int len) {
+        return len & 0xffffffffL;
```
hczedik, line 688: We should consider the special -1 case here for the conversion:
`if( length == -1 ) return -1;`

SamuelImriska:
> Yes, I think that makes sense, @SamuelImriska:
> 
> Change the field "length" in DicomInputStream from int to long.
> 
> Maybe keep the existing method length() as an int, but mark it as deprecated, and suggest using the new longLength() method.
> 
> Same would be nice for BulkData, but that is a bit more tricky, as it is a Serializable so that could break serialized objects if not done carefully.
> 
> What do you think, @gunterze. Go for a small change here, or a bigger changeset?

But the writeObject from Serialization does not write the length at all. So should not be a problem

gunterze:
Fixed by https://github.com/dcm4che/dcm4che/commit/28a3a5b2a02eb96d920d80dab7bc670924a0f7b9

PR #1290, opened 2023-01-24
Support reading 2.1GB+ DICOM files

Thaliant:
This fork makes what I hope is a non-invasive change allowing DCM4CHE to read DICOM files that are 2.1GB or larger (e.g. tomography) by leveraging a recent improvement that allows `BulkData` to manage `long`-length data. When `DicomInputStream` encounters a tag with a length greater than the maximum allowed `byte[]` length, it injects a `BulkData` instance into the `Attributes` instead of throwing an exception that the tag value is too long.

I've tested this change through our own code although the easiest way to see it in action is to use `dcm2jpg` with a giant tomography file. Rather than throwing an exception, it produces a JPG of the first frame of the file. You can also see that `dcmdump` is able to dump the contents of the file including the first bit of `PixelData` that previously would cause an exception. I didn't exhaustively test the other tools but assuming `BulkData` is handled properly, this should _just work._

Appreciate the consideration. Thanks!

### DIFF
--- a/dcm4che-core/src/main/java/org/dcm4che3/io/DicomInputStream.java
+++ b/dcm4che-core/src/main/java/org/dcm4che3/io/DicomInputStream.java
@@ -729,6 +729,8 @@ public class DicomInputStream extends FilterInputStream
         } else if (length == BulkData.MAGIC_LEN
                 && super.in instanceof ObjectInputStream) {
             attrs.setValue(tag, vr, deserializeBulkData((ObjectInputStream) super.in));
+        } else if (length > Integer.MAX_VALUE) {
+            attrs.setValue(tag, vr, bulkDataCreator.createBulkData(this));
         } else if (includeBulkDataURI) {
             attrs.setValue(tag, vr, bulkDataCreator.createBulkData(this));
         } else {
--- a/dcm4che-core/src/test/java/org/dcm4che3/io/DicomInputStreamTest.java
+++ b/dcm4che-core/src/test/java/org/dcm4che3/io/DicomInputStreamTest.java
@@ -86,6 +86,85 @@ public class DicomInputStreamTest {
     }
 
     @Test
+    public void testPixelDataLarger2GiB() throws Exception {
+
+        // Two temporary files - one to hold 2.1+Gib of PixelData and one to
+        // hold the encoded DICOM file containing that PixelData. The goal of
+        // this test is to ensure that DicomInputStream doesn't throw an
+        // exception when it receives PixelData in excess of Integer.MAX_VALUE.
+        final File pixelDataFile = File.createTempFile("dcm3che-", "-3gib-pixel-data");
+        final File dicomFile = File.createTempFile("dcm3che-", "-3gib-dicom");
+
+        try {
+
+            // This will hold the total number of pixel data written.
+            long totalPixelDataBytes = 0;
+
+            // Fill a buffer with random data. We're going to verify that
+            // the data we read back in matches this buffer.
+            final byte[] buffer = new byte[1024 * 1024 * 18];
+            for (int i = 0; i < buffer.length; ++i)
+                buffer[i] = (byte) Math.round(Math.random() * 128);
+
+            // This test writes in excess of 2.5GiB of data for good measure.
+            try (final FileOutputStream outputStream = new FileOutputStream(pixelDataFile)) {
+                while (totalPixelDataBytes < 2508221998L) {
+                    outputStream.write(buffer);
+                    totalPixelDataBytes += buffer.length;
+                }
+            }
+
+            final String patientId = "ABC123";
+
+            // Need a DICOM file with a regular attribute to verify that we
+            // can read the resulting file back.
+            final Attributes attributes = new Attributes();
+            attributes.setString(Tag.PatientID, VR.LO, patientId);
+            attributes.setValue(Tag.PixelData, VR.OB, new BulkData(pixelDataFile.toURI().toString(), 0, totalPixelDataBytes, false));
+
+            final Attributes fmi = new Attributes();
+            fmi.setString(Tag.FileMetaInformationGroupLength, VR.UL, "1");
+            fmi.setString(Tag.MediaStorageSOPClassUID, VR.UI, UID.ComputedRadiographyImageStorage);
+            fmi.setString(Tag.TransferSyntaxUID, VR.UI, UID.ExplicitVRLittleEndian);
+
+            try (final DicomOutputStream outputStream = new DicomOutputStream(dicomFile)) {
+                outputStream.writeDataset(fmi, attributes);
+            }
+
+            // Now read the resulting file back in. Verify that both the regular
+            // tags and the giant PixelData is valid.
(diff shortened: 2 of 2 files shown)


Thaliant:
Hello @gunterze ! Hoping that you'll consider this PR at your earliest convenience. Would be really helpful to have it in the next release of DCM4CHE if this change meets your expectations. Cheers!

Thaliant:
Hello @vrindanayak ! I see you are also an active contributor to the project. Your thoughts on this PR would be extremely helpful. It would be very useful to get this rather simple change into the next release of DCM4CHE. I believe there is a growing number of library users that are bumping into limitations when trying to process giant Mammography studies with DCM4CHE and this PR addresses the issue. Thanks in advance!

gunterze:
You may set `DicomInputStream.setIncludeBulkData(IncludeBulkData.URI)` before parsing the dataset anyway.
`dcm2jpg` relies on `DicomImageReader` which stops parsing the dataset and allocating `byte[]` for attribute value **before** the pixel data! So I don't see, why `dcm2jpg` should fail rendering a single frame of large multi-frame images > 2GiB without your proposed change.

Thaliant:
@gunterze Thank you for the reply! The issue with your suggestion is that in the tomography files we're processing are _not_ multi-frame. They're single images with pixel data in excess of 2.1GB. Here's code to illustrate the problem:

```java
try (final DicomInputStream inputStream = new DicomInputStream(new File("giant.dcm"))) {
    inputStream.setIncludeBulkData(DicomInputStream.IncludeBulkData.URI);
    final Attributes attributes = inputStream.readDataset();
    System.out.println(attributes.getString(Tag.PatientID));
} catch (final Exception ex) {
    ex.printStackTrace();
}
```

This yields the following exception:

```
org.dcm4che3.io.DicomStreamException: 0x7FE00010 tag value too large, must be less than 2Gib
	at org.dcm4che3.io.DicomInputStream.tagValueTooLargeException(DicomInputStream.java:500)
	at org.dcm4che3.io.DicomInputStream.readHeader(DicomInputStream.java:539)
	at org.dcm4che3.io.DicomInputStream.readAttributes(DicomInputStream.java:661)
```

To be clear, our use case is that we need to be able to load the entire dataset, _including pixel data_. Our system is  manipulating the DICOM and writing out a new, separate file. The goal of my PR is change the behavior from throwing an exception to loading in excessive-sized `PixelData` as `BulkData` which works extremely well - the pixel data isn't actually loaded into memory (which is an awesome bonus) and we can fully manipulate and then write out the entire DICOM quickly and efficiently. 

@sander4318 I will get working on the unit test you requested!

Thaliant:
@sander4318 Unit test committed! It generates 2.5GB of fake PixelData, shoves it into a DICOM file, reads it back in and verifies that `DicomInputStream` can read it back in seamlessly. Cheers!

wirbeiagfa:
I think the root cause here is that the current BulkDataDescriptors do not allow describing large values (>2GB) correctly, as already noted here: https://github.com/dcm4che/dcm4che/commit/28a3a5b2a02eb96d920d80dab7bc670924a0f7b9#r92301125

hczedik:
I think the root cause here is that the current BulkDataDescriptors do not allow describing large values (>2GB) correctly, as already noted here: https://github.com/dcm4che/dcm4che/commit/28a3a5b2a02eb96d920d80dab7bc670924a0f7b9#r92301125

gunterze:
Could not reproduce:
```java
    @Test
    public void testPixelDataLarger2GiB() throws Exception {
        File dicomFile = File.createTempFile("dcm4che-3gib-", ".dcm");
        dicomFile.deleteOnExit();
        byte[] PIXELDATA_HEADER = {
                (byte) 0xE0, 0x7F, 0x10, 0x00, 'O', 'W', 0x00, 0x00, 0x00, 0x00, (byte) 0x80, (byte) 0x94
        };

        // Fill a buffer with random data. We're going to verify that
        // the data we read back in matches this buffer.
        byte[] buffer = new byte[1024 * 1024 * 18];
        for (int i = 0; i < buffer.length; ++i)
            buffer[i] = (byte) Math.round(Math.random() * 128);

        Attributes fmi = new Attributes();
        fmi.setInt(Tag.FileMetaInformationVersion, VR.OB, 1);
        fmi.setString(Tag.TransferSyntaxUID, VR.UI, UID.ExplicitVRLittleEndian);
        Attributes dataset = new Attributes(1);
        dataset.setString(Tag.PatientID, VR.LO, "ABC123");
        try (DicomOutputStream dos = new DicomOutputStream(dicomFile)) {
            dos.writeDataset(fmi, dataset);
            dos.write(PIXELDATA_HEADER);
            for (int i = 0; i < 132; i++) dos.write(buffer);
        }
        try (DicomInputStream dis = new DicomInputStream(dicomFile)) {
            dis.setIncludeBulkData(IncludeBulkData.URI);
            dataset = dis.readDataset();
        }
        Object pixelData = dataset.getValue(Tag.PixelData);
        assertTrue(pixelData instanceof BulkData);

        // We're not going to compare the whole thing - just the first
        // segment should match byte-for-byte. We assume the rest of
        // the data will be correct.
        try (InputStream is = ((BulkData) pixelData).openStream()) {
            byte[] buffer2 = new byte[buffer.length];
            is.read(buffer2);
            assertArrayEquals(buffer, buffer2);
        }
   ```
    succeeds.

**Comment by Thaliant** [comment]
@gunterze When I back out my fix in `DicomInputStream` and run `./mvnw install` the test you provided _does run successfully_ as you said. However, my test creates a giant DICOM file consistent with the one we received from our customer that fails. With my change backed out, my test fails and throws the `value too large` exception. Is it possible that your test and my test write the `PixelData` tag header differently? One triggers the exception, the other does not.

**Comment by Thaliant** [comment]
@gunterze I'm trying to debug locally with the customer DICOM file that fails to see if I can figure out why the `isBulkData()` call is failing for our particular dataset whereas it doesn't happen in yours. I believe that @hczedik is right that the real root problem is there. I'll see if I can figure out what's special about this particular dataset.

**Comment by gunterze** [comment]
I also cannot reproduce the issue with your test provided in https://github.com/dcm4che/dcm4che/pull/1290/commits/d396d75747a78ca36a54e01989cc88ff5bca1347

just inserting
```
                inputStream.setIncludeBulkData(IncludeBulkData.URI);
```
before
```
                final Attributes _attributes = inputStream.readDataset();
```

Thaliant:
@gunterze @sander4318 @hczedik **I apologize.** It appears that the issue I was trying to fix is no longer an issue in DCM4CHE 5.29.2. 

DCM4CHE 5.29.0 (which is what I believe my original fork was based on) exhibits the `value too large` exception when trying to read the DICOM file whereas DCM4CHE 5.29.2 with the `IncludeBulkData.URI` option enabled reads it without problem. I thought I had merged 5.29.2 into my fork before I made my test yesterday but it appears I was still running locally with the 5.29.0 files.

Thank you for the discussion and feedback.  Sorry for what has turned out to be a wild goose chase. I will cancel this PR.

Thaliant:
PR is no longer applicable in DCM4CHE 5.29.2 which no longer exhibits the issue we were trying to address.

Issue #1323, opened 2023-07-06
Transcoder: java.lang.ClassCastException


adrianog3:
Affected Library: dcm4che-imageio

When trying to transcode a dicom file with transfer syntax 1.2.840.10008.1.2.4.70 (JPEGLosslessSV1) to 1.2.840.10008.1.2.4.80 (JPEGLSLossless) the operation is successful, but when trying to transcode the output file to the original transfer syntax the following error happens:

`java.lang.ClassCastException: java.awt.image.DataBufferByte cannot be cast to java.awt.image.DataBufferShort
	at org.dcm4che3.image.BufferedImageUtils.convertShortsToBytes(BufferedImageUtils.java:169) ~[dcm4che-image-5.24.0.jar:5.24.0]
	at org.dcm4che3.imageio.codec.Transcoder.compressPixelData(Transcoder.java:579) ~[dcm4che-imageio-5.24.0.jar:5.24.0]
	at org.dcm4che3.imageio.codec.Transcoder.processPixelData(Transcoder.java:507) ~[dcm4che-imageio-5.24.0.jar:5.24.0]
	at org.dcm4che3.imageio.codec.Transcoder.access$800(Transcoder.java:71) ~[dcm4che-imageio-5.24.0.jar:5.24.0]
	at org.dcm4che3.imageio.codec.Transcoder$1.readValue(Transcoder.java:464) ~[dcm4che-imageio-5.24.0.jar:5.24.0]
	at org.dcm4che3.io.DicomInputStream.readAttributes(DicomInputStream.java:576) ~[dcm4che-core-5.24.0.jar:5.24.0]
	at org.dcm4che3.io.DicomInputStream.readAllAttributes(DicomInputStream.java:483) ~[dcm4che-core-5.24.0.jar:5.24.0]
	at org.dcm4che3.imageio.codec.Transcoder.transcode(Transcoder.java:438) ~[dcm4che-imageio-5.24.0.jar:5.24.0]
	at org.example.Application.transcode(Application.java:63) [classes/:?]
	at org.example.Application.main(Application.java:46) [classes/:?]`

In my tests, only a few files failed to revert the transfer syntax to the original value.

To reproduce the error i created a repository with instructions and a sample: https://github.com/adrianog3/dcm4che-transcode-issue

The error does not depend on the operating system.

adrianog3:
Hi guys, i found the solution and open a PR for code review. 

[...]

PR #1324, opened 2023-07-07
fix: transcode between JPEGLosslessSV1 and JPEGLSLossless

adrianog3:
When data from source BufferedImage was byte array the code fails when performed cast.

The proposed solution is to check if the data from source BufferedImage is already in bytes, and copy it directly.


jhonatasfender (approved):

adrianog3:
The related issue is: https://github.com/dcm4che/dcm4che/issues/1323

lucianomukudai (approved):

romilson-junio (approved):

nroduit:
If you're interested in using Java APIs directly for transcoding, then take a look at this [library](https://github.com/nroduit/weasis-dicom-tools), which is based on dcm4che3 but where image management has been completely redesigned and no longer uses imageio (except to register the new native code).

The above-mentioned problem is not present when using the transcoding API.

Here are some new features or behaviors:

-  Less complex code for reading, writing and tanscoding images (remove Java reflection in stream classes)
-  Only one pixel model (interleaved mode) returned by the image reader. Sub-sampled models are automatically convert to normal interleaved image. This eliminates the complexity at the end of the chain.
-  Keep pixel data as a native object when reading, writing, converting, resizing, and all the image processing operations (except for the last operation of display, needs to convert into Bufferedimage)
-  Return raw pixel data (byte, short, ushort, int, float, double) without applying modality, voi and presentation LUT
-  Large API for image processing and conversion
-  ImageDescriptor contains the modality and voi LUT, and the overlay data.
-  ImageReaderDescriptor has two sub-classes: 1) DicomFileInputStream: handle DICOM file 2) BytesWithImageDescriptor: handle image content in bytes
-  ImageReader contains getLazyPlanarImages which allows to apply several processing functions without having pixel in memory (image is read only when calling PlanarImage.get()). This is particularly interesting for handling multiframe separately and not loading all frames in memory.

adrianog3:
Hi @nroduit, thank you for the answer.

I use the transcoding feature on many parts of the system, so migrating to [weasis-dicom-tools](https://github.com/nroduit/weasis-dicom-tools) would take a long time as it requires doing regression testing.

Given the context, I'd like to know if my PR might be the solution to the problem until I migrate to weasis-dicom-tools, or if in fact I should migrate now.

nroduit:
Yes, that seems to fix the issue for this particular case. I don't have permission to merge.

adrianog3:
I don't know if I should be asking this question here, but i'm trying to evaluate the use of the [weasis-dicom-tools](https://github.com/nroduit/weasis-dicom-tools) library and i'm having trouble importing the latest version: **5.29.2.2**.

Adding the dependency:

`<dependency>
    <groupId>org.weasis</groupId>
    <artifactId>weasis-dicom-tools</artifactId>
    <version>5.19.1</version>
</dependency>`

And this repository:

`<repository>
    <id>mvn-repo-master</id>
    <url>https://raw.github.com/nroduit/mvn-repo/master/</url>
    <snapshots>
        <enabled>false</enabled>
    </snapshots>
</repository>`

Only version **5.19.1** is available.

Is there another repository I should add to get access to version **5.29.2.2**?

image

adrianog3:
Thank you very much @nroduit, 

I analyzed the [weasis-dicom-tools](https://github.com/nroduit/weasis-dicom-tools) code and concluded that the Transcoder does not allow to provide InputStream and OutputStream.

These parameters is very important for my use case, because I do the reading dcm files in a certain transfer syntax, and I send it through the http protocol using another transfer syntax, so generating files is unnecessary.

Can I implement this improvement and open a pull request?

adrianog3:
Thank you very much @nroduit, 

I analyzed the [weasis-dicom-tools](https://github.com/nroduit/weasis-dicom-tools) code and concluded that the Transcoder does not allow to provide **OutputStream**.

This parameter is very important for my use case, because I do the reading dcm files in a certain transfer syntax, and I send it through the http protocol using another transfer syntax, so generating files is unnecessary.

Is it possible to implement this improvement? If you allow me I can carry out this implementation.

adrianog3:
Can anyone review and merge this PR?

PR #1348, opened 2023-09-19
Split date/time related tests into own file

hczedik:
Split the existing date/time related tests into an own test class.

Reasoning: I want to extend these tests in an upcoming PR (for supporting modern java.time types).
It makes sense to have all the date/time related Attributes tests bundled within one file to keep a better overview.

@gunterze: It would be helpful if you can review/merge this PR quickly. (And also the other prerequisite PR: https://github.com/dcm4che/dcm4che/pull/1347)
This will ease the review of the upcoming PR, because you can then focus on the actual functional changes without such pure refactorings.


gunterze (approved):

hczedik:
Thanks, @gunterze for the fast review and merge!

PR #1351, opened 2023-09-26
Support modern java.time types for getters (first draft)

hczedik:
Add a getTemporal(.) method to Attributes which returns one of the modern java.time types (ZonedDateTime, LocalDateTime).

Motivation: The current timezone handling using the legacy Date types is a source of confusion and bugs, because it is unclear on whether a timezone corrected date/time or a local date/zime is returned.

@gunterze: This is a draft containing some TODOs. I just want to get your general feedback if you agree with the general design and whether I should continue working on this. Thanks!


dcm4che-core/src/main/java/org/dcm4che3/data/Attributes.java
```diff
@@ -1671,6 +1848,16 @@ public void setDefaultTimeZone(TimeZone tz) {
...
     }
 
+    public ZoneId getDefaultZoneId() {
+        if (defaultTimeZone != null)
+            return defaultTimeZone.toZoneId();
+
+        if (parent != null)
+            return parent.getDefaultZoneId();
+
+        return null;
```
hczedik, line 1858: Note the different behavior here compared to getDefaultTimezone: It returns null in case nothing is defined and does not fall back to the system default.

gunterze (commented):
Would prefer to do **not** consider the "corresponding"  TM/DA attribute for DA/TM attributes.

Instead I propose to return

| VR | &ZZXX | TimeZone | DST | Class           |
|----|-------|----------|-----|-----------------|
| DT | no    | no       | n/a | `LocalDateTime` |
| DT | yes   | *        | *   | `OffsetDateTime` |
| DT | no    | yes      | no  | `OffsetDateTime` |
| DT | no    | yes      | yes | `ZonedDateTime` |
| DA | n/a   | *        | *   | `LocalDate`     |
| TM | n/a | no       | n/a | `LocalTime`     |
| TM | n/a | yes      | no  | `OffsetTime`    |
| TM | n/a | yes      | yes | `LocalTime`     |

for `getTemporal(int tag)`, where TimeZone fallbacks  to defaultTimeZone,

And provide also `getTemporal(long tag)` returning

| DA  | TM  | TimeZone | DST | Class            |
|-----|-----|----------|-----|------------------|
| no  | *   | *        | *   | `null`           |
|  yes   | no  | *        | *   | `LocalDate` |
| yes | yes | no       | n/a | `LocalDateTime`  |
|  yes   | yes | *        | *   | `OffsetDateTime` |
| yes    | yes | yes      | no  | `OffsetDateTime` |
|  yes   | yes | yes      | yes | `ZonedDateTime`  |

hczedik:
Hi @gunterze ,

thanks a lot for the detailled feedback on this one. Sorry for the delay now on my side, I am busy with some other stuff at the moment. But I will definitely follow-up on this PR.

Thanks again!

hczedik:
Hi @gunterze,

I finally found some time to follow up on this one here.

Thanks again for your detailled feedback.

I have now started adapting the implementation based on your feedback, but I would like to make some suggestions to simplify the API:

For `getTemporal(int)`:

| VR | &ZZXX | TimeZone | DST | Class  |
|--|--|--|--|--|
| DT | yes | * | * | `ZonedDateTime` |
| DT | no | no | n/a | `LocalDateTime` |
| DT | no | yes | * | `ZonedDateTime` |
| DA | n/a | * | * | `LocalDate` |
| TM | n/a | * | * | `LocalTime` |

Reasoning: I would return ZonedDateTime always and never OffsetDateTime. While OffsetDateTime is technically more accurate, I think from a user perspective the method will be simpler to understand if there are not too many different return types.
ZonedDateTime can perfectly represent any OffsetDateTime without loss of information. In case a user is really interested if it is a timezone with DST or a fixed offset, the `getZone().getRules().isFixedOffset()` method can be used on the returned temporal.

I would also avoid the OffsetTime type (for TM tags) and always simply use LocalTime, again to keep it simple. If a user is really interested in the offset, they can access it separately via the getZoneId() method.


Consistently, I would propose for the getTemporal(long) method:

| DA | TM | TimeZone | DST | Class |
|--|--|--|--|--|
| no | * | * | * | `null` |
| yes | no | * | * | `LocalDate` |
| yes | yes | no | n/a | `LocalDateTime`  |
| yes | yes | * | * | `ZonedDateTime` |

Is that acceptable for you?

hczedik:
@gunterze, did you already have time to take a look at my new draft?

Thank you!

dcm4che-core/src/main/java/org/dcm4che3/data/Attributes.java
```diff
@@ -1701,6 +1836,31 @@ public TimeZone getTimeZone() {
         return tz;
      }
 
+    public ZoneId getZoneId() {
+        // TODO we might want to store the ZoneId, instead of converting every time here
```
thomasdewaelheyns, line 1840: The ZoneId is cached inside the TimeZone object itself:

```
public ZoneId toZoneId() {
        ZoneId zId = zoneId;
        if (zId == null) {
            zoneId = zId = toZoneId0();
        }
        return zId;
    }
```

hczedik: Good find, thank you!

PR #1398, opened 2024-02-13
IEI-299706 Removed whitespace in the application name used in MessageBuilder.writeHeader method

stephenfrederick:
Remove whitespace from the header used for syslog to avoid compatibility problems with 3rd party solutions that check the syslog header for validity.

PR #1399, opened 2024-02-14
Removed whitespace in the application name used in MessageBuilder.writeHeader method

stephenfrederick:
Remove whitespace from the header used for syslog to avoid compatibility problems with 3rd party solutions that check the syslog header for validity.


njiddasalifu (approved):
LGTM

hczedik:
Hi @gunterze, can you provide some feedback here if the general approach is acceptable?
We'll need to merge the linked https://github.com/dcm4che/dcm4che/pull/1398 PR, and of course I would prefer that the approach on the generic-config and master version stay in sync.

gunterze:
Fixed by https://github.com/dcm4che/dcm4che/issues/1406

PR #1418, opened 2024-04-06
Fix #1417

### LINKED ISSUE #1417: Build fails on Windows
When running the build on Windows it fails with the following error:

```
[ERROR] Failed to execute goal org.codehaus.mojo:build-helper-maven-plugin:3.4.0:regex-property (resolve weasis.opencv.native.version - property) on project dcm4che-assembly: No match to regex '.*libopencv_java\/(.*)\/libopencv_java.*' found in 'D:\Users\axliy\.m2\repository\org\weasis\thirdparty\org\opencv\libopencv_java\4.9.0-dcm\libopencv_java-4.9.0-dcm-linux-x86-64.so'.
```

**To Reproduce**
* Run `mvnw clean install`

Operating system: Windows 10

@nroduit: This is due to https://github.com/dcm4che/dcm4che/pull/1410. Could you please take a look at this?

nroduit:
See #1417

### DIFF
--- a/dcm4che-assembly/pom.xml
+++ b/dcm4che-assembly/pom.xml
@@ -163,7 +163,7 @@
               <name>weasis.opencv.native.version</name>
               <!--suppress UnresolvedMavenProperty -->
               <value>${org.weasis.thirdparty.org.opencv:libopencv_java:so:linux-x86-64}</value>
-              <regex>.*libopencv_java\/(.*)\/libopencv_java.*</regex>
+              <regex>.*libopencv_java[\/\\](.*)[\/\\]libopencv_java.*</regex>
               <replacement>$1</replacement>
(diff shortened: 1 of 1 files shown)


hczedik:
Hi @nroduit ,

thanks a lot for taking a look into this.

I tried out your fix here, and it now seems to fail with:

[ERROR] Failed to execute goal org.codehaus.mojo:build-helper-maven-plugin:3.4.0:regex-property (resolve weasis.opencv.native.version - property) on project dcm4che-assembly: The parameters 'value' for goal org.codehaus.mojo:build-helper-maven-plugin:3.4.0:regex-property are missing or invalid

hczedik:
Ah, sorry, I was using an older mvn version by accident.
If I use the correct version (the one specified by the Maven wrapper mvnw script), then it works now.

Thanks a lot!

PR #1444, opened 2024-07-25
Patch to support JPEG2000 images with 16 bits allocated but 8 bits stored

georgsam91:
Patch to DCM4CHE to allow support for JPEG2000 images that are 16 bits allocated but only 8 bits stored. Currently, these images are not correctly decompressed by DCM4CHE. DCM4CHE will remove some of the components of the image. This implements a patch in DCM4CHE to convert the 
image to 8 bits allocated and 8 bits stored and removes the empty leading 00 bits. After this conversion DCM4CHE correctly decompresses the image and the pixel data checksum also passes. Since we are changing the BitsAllocated tag this will also update the BitsAllocated in the 
ImageSetMetadata.

As an example in hex form the original images appear in the following form with each two-character sequence representing a byte (i.e. 8 bits) and each four-character sequence representing a short (i.e. 16 bits). For example, 001E represents 16-bits or a single pixel value for a single 
image component. This will be converted to 8 bits 1E by stripping off leading 00s which do not contain any data.

Original Hex

001E0030 0048001E 00370048

After Hex

1E30 481E 3748

### DIFF
--- a/dcm4che-imageio/src/main/java/org/dcm4che3/imageio/codec/Transcoder.java
+++ b/dcm4che-imageio/src/main/java/org/dcm4che3/imageio/codec/Transcoder.java
@@ -836,7 +836,13 @@ public class Transcoder implements Closeable {
                 write(sm, ((DataBufferByte) db).getBankData());
                 break;
             case DataBuffer.TYPE_USHORT:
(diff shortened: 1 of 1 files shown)


georgsam91:
Seeking a review for this change to support JPEG2000 images

PR #1449, opened 2024-08-26
IEI-320939

alsirhc:
Changes to 
- handle InsufficientCapacityException thrown by AuditLogger using default audit exception handler, which adds failed message to queue 
- allow modification ring buffer size and thread pool core size for testing and potential (temporary) mitigation

### DIFF
--- a/dcm4che-net-audit/src/main/java/org/dcm4che3/net/audit/AuditLogger.java
+++ b/dcm4che-net-audit/src/main/java/org/dcm4che3/net/audit/AuditLogger.java
@@ -106,7 +106,13 @@ public class AuditLogger extends DeviceExtension {
 
     private static final int MSG_PROMPT_LEN = 8192;
 
+    private static final int DEFAULT_RING_BUFFER_SIZE = 1024;
+
+    private static final int MINIMUM_RING_BUFFER_SIZE = 2;
+
     private static Logger LOG = LoggerFactory.getLogger(AuditLogger.class);
+
+    private static AuditMessageExceptionHandler messageExceptionHandler;
 
     public enum Facility {
         kern,            // (0) -- kernel messages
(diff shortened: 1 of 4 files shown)


hczedik:
Removing myself as reviewer, as I am not familiar with the auditing code, and the changes will not have impact on the IA code.

dcm4che-net-audit/src/main/java/org/dcm4che3/net/audit/AuditLogger.java
```diff
@@ -743,6 +756,61 @@ public void setRetryInterval(int interval) {
...
+     * @param value value to check as power of two
+     * @return number that's the smallest power of two greater than or equal to number
+     */
+    private static int getPowerOfTwo(int value) {
+        if (value <= 1) {
+            return 2;
+        };
+
+        int next = Integer.highestOneBit(value);
+        if (value == next) {
```
stephenfrederick, line 805: You don't really need this check. Just remove the if clause entirely and use the following return statement.

dcm4che-net-audit/src/main/java/org/dcm4che3/net/audit/SyslogProtocolHandler.java
```diff
@@ -103,7 +104,7 @@ private static int getThreadPoolCoreSize() {
                 System.getProperty(AUDIT_THREAD_POOL_CORE_SIZE, String.valueOf(MINIMUM_NUMBER_OF_THREADS)));
         threadPoolSize = Math.max(Runtime.getRuntime()
                 .availableProcessors(), threadPoolSize);
-        // LOG.warn("Using audit thread pool core size {}", threadPoolSize);
+        SyslogLogger.LOG.debug("Audit core thread pool size: {}", threadPoolSize);
```
stephenfrederick, line 107: Just use LOG here?

alsirhc: Believe me, tried that but it results in an exception thrown, perhaps because of something _special_ about how the class is instantiated. While ugly, creating this private logging class worked.

stephenfrederick (approved):

Issue #1451, opened 2024-09-05
6.2.2 Unknown (UN) Value Representation implementation that attempts to read a UN as a SQ. fails for a Dicom ingested in the past by storage using dcm4che version2.0.25


bdasilva-ims:
6.2.2 Unknown (UN) Value Representation implementation that attempts to read a UN as a SQ. fails for a Dicom ingested in the past by storage using dcm4che version 2.0.25

1: additional test for DicomInputStreamTest
`    @Test
    public void uncorrectableUNSequenceVR() throws IOException {
        byte[] b = {  0x08,  0, 0x11, 0x30,
                      'U', 'N', 0, 0,
                      0x48,  0, 0, 0,
                      0x10, 0x48, (byte) 0xDD, 0x1D,
                      0x22, 0x02,        0x11, 0x48,
                      (byte) 0xE6, (byte) 0xB7, (byte) 0xCD, (byte) 0xCD, (byte) 0xED, (byte) 0xFD, (byte) 0xEE, (byte) 0xB7,
                      (byte) 0xEC, (byte) 0xD3, (byte) 0xDC, (byte) 0x99, (byte) 0xE5, (byte) 0xC9, (byte) 0xDE, (byte) 0x99,
                      (byte) 0xEC, (byte) 0xCC, (byte) 0xDD, (byte) 0x81, (byte) 0xE5, (byte) 0xCC, (byte) 0xC0, (byte) 0x85,
                      (byte) 0xE4, (byte) 0xCF, (byte) 0xDE, (byte) 0x82, (byte) 0xE9, (byte) 0xCB, (byte) 0xD8, (byte) 0x80,
                      (byte) 0xEF, (byte) 0xD3, (byte) 0xDF, (byte) 0x82, (byte) 0xE9, (byte) 0xC9, (byte) 0xDF, (byte) 0x81,
                      (byte) 0xE8, (byte) 0xC5, (byte) 0xDD, (byte) 0x83, (byte) 0xF3, (byte) 0xC9, (byte) 0xDB, (byte) 0x8E,
                      (byte) 0xEB, (byte) 0xD3, (byte) 0xDC, (byte) 0x80, (byte) 0xE8, (byte) 0xCE, (byte) 0xDD, (byte) 0x81,
                             0x10,        0x48, (byte) 0xD0,        0x1D, (byte) 0xDD, (byte) 0xFD, (byte) 0xEE, (byte) 0xB7};

        Attributes attrs;
[...]

PR #1452, opened 2024-09-05
6.2.2 Unknown (UN) Value Representation implementation that attempts to read an UN as a SQ. fails for a Dicom ingested in the past by storage using dcm4che version 2.0.25 #1451

bdasilva-ims:
6.2.2 Unknown (UN) Value Representation implementation that attempts to read an UN as a SQ. fails for a Dicom ingested in the past by storage using dcm4che version 2.0.25 #1451

proposed fix is to read ahead 4 bytes to check if any of the tags ( Tag.Item, Tag.ItemDelimitationItem, Tag.SequenceDelimitationItem ) are present, if they do readAttributes goes ahead and reads the UN as a SQ
otherwise it leaves the VR as UN

### DIFF
--- a/dcm4che-core/src/main/java/org/dcm4che3/io/DicomInputStream.java
+++ b/dcm4che-core/src/main/java/org/dcm4che3/io/DicomInputStream.java
@@ -764,8 +764,26 @@ public class DicomInputStream extends FilterInputStream
                             if (vr == VR.UN && length == UNDEFINED_LENGTH)
                                 vr = VR.SQ; // assumes UN with undefined length are SQ,
                             // will fail on UN fragments!
+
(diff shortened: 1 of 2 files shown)


dmurphy-ambra:
@gunterze Would it be possible to merge this fix? We have a client that uses dcm4che 2.0 in a very old PACS implementation, and they generate DICOM tag (0010,0011) as VR UN.  Now that (0010,0011) is officially part of the DICOM standard, dcm4che3 attempts to convert the UN to SQ. When dcm4che3 parses this tag as SQ, it hits the same issue addressed by the PR - an EOFException.  We've tested this fix and verified it resolves the issue.  Would really appreciate if this could be reviewed. Thanks.

PR #1493, opened 2025-04-29
IEI-352818 wired required ConfigurablePropety into schema code and adapted code to use

msiemczyk:
- wired required attribute into schema propagation so UI can use it
- changed enum schema generation code to use required property for allowing null instead of assuming that specified default value means no nulls allowed and modified properties that relied on this assumption to required to keep the UI consistent
- made TransferCapability fields required that throws exception when unset and are truly required


dcm4che-conf/dcm4che-conf-core/src/test/java/org/dcm4che3/conf/core/tests/adapters/DefaultConfigTypeAdaptersTest.java
```diff
@@ -0,0 +1,117 @@
...
+	    private EnumTypeAdapter adapter = new EnumTypeAdapter();
+	    
+		@Test
+		void getSchema_ReturnsCorrectlyPopulatedMap_GivenValidEnumConfigProperty() {
+
+			List<ConfigProperty> configurableFields = ConfigReflection.getAllConfigurableFields(TestEnumClass.class);
+
+			final Map<String, Object> actualSchema = adapter.getSchema(configurableFields.get(0), null);
+			
+	        Map<String, Object> expectedSchema = new HashMap<>();
```
stephenfrederick, line 38: Spacing looks off in this file.

stephenfrederick (approved):

dcm4che-conf/dcm4che-conf-core/src/test/java/org/dcm4che3/conf/core/tests/adapters/DefaultConfigTypeAdaptersTest.java
msiemczyk: will fix it, I had some issues with tabs vs spaces in IDE

PR #1514, opened 2025-07-10
WADO: 1/0 Windowing creates single value LUT with wrong value (0x00 i…

fwiller:
…nstead of 0xff)


gunterze:
Fixed by https://github.com/dcm4che/dcm4che/commit/b55d1da44bef13208fa0ee46b0aea11d01a80807

PR #1515, opened 2025-07-15
IEI-190865 Align EI Implementation Class UID with Agfa's UID root

KurtHuber69:
Align EI Implementation Class UID with Agfa's UID root- association negotiation

Update DICOM Implementation Class UID uses in association negotiation to use the new EI Implementation Class UID (based on the new Agfa UID root): 1.3.51.0.42.1.1


hczedik:
Closing because of inactivity.

PR #1517, opened 2025-07-25
IEI-355864 Study Display and Study Export (from LTA) fails when individual composite tag is added to the Instance Level DICOM attrs filter

KurtHuber69:
- Change the 'updateRecursive' function in Attributes class so that it keeps track of modified attributes so that this function can be used when determining explicitly coerced attributes.

### DIFF
--- a/dcm4che-core/src/main/java/org/dcm4che3/data/Attributes.java
+++ b/dcm4che-core/src/main/java/org/dcm4che3/data/Attributes.java
@@ -2069,7 +2069,19 @@ public class Attributes implements Serializable {
      * @return <tt>true</tt> if one ore more attribute are added or
      *          overwritten with a different value
      */
-    public boolean updateRecursive (Attributes other) {
+    public boolean updateRecursive(Attributes other) {
+    	return updateRecursive(other, null);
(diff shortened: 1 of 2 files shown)


andrew-cowan (approved):
Looks good:  being able to track modified attributes is critical functionality which will be good to have in the generic-config branch

KurtHuber69:
@hczedik The 'updateRecursive' function is only in the generic-config branch, so these changes are only applied in this branch.
If we wanted to copy the changes to master, then the previous issue where the 'updateRecursive' function was added would need to be merged first, however, the master branch uses 'UpdatePolicy' whereas generic-config does not and so this makes the code for this API quite different.

hczedik:
Ah, I see. We should try to align with the functionality on the master branch as much as possible.

In my opinion, we should stop using updateRecursive in our code, and switch to using the update methods available on master (pull them to generic-config if necessary). That one already seems to support tracking the modified attributes. So it seems to be a good fit for what you need.

hczedik:
@gunterze is this correct? Would it make sense to add this functionality to master?

PR #1536, opened 2025-10-02
IEI-354094 Override missing toBytes(long l,..) in BYTE BinaryValueType to fix XML parsing when VR of tag is OB

emyjacob:
Added missing method toBytes(long l,..) in BYTE BinaryValueType


hczedik:
I do not see this method on the master branch. Why is it needed on generic-config but not master?

hczedik:
Also, can you cover this with a unit test?

hczedik (approved):

PR #1551, opened 2025-11-14
Fix image: When Tag.ICCProfile has invalid value, loading of Image fails

ivosek:
**Problem:**
Data from `SIEMENS MAGNETOM Vida` contains invalid ICCProfile tag value 
```
(0028,2000) OB 02\03\04\05                              #   4, 1 ICCProfile
```
and in this case dcm4che-image fails to load image.

**Exception:**
```
Caused by: java.lang.IllegalArgumentException: Invalid ICC Profile Data
	at java.desktop/sun.java2d.cmm.ProfileDataVerifier.verify(ProfileDataVerifier.java:42)
	at java.desktop/java.awt.color.ICC_Profile.getInstance(ICC_Profile.java:789)
	at org.dcm4che3.image.ICCProfile.colorSpaceFactoryOf(ICCProfile.java:87)
	at org.dcm4che3.imageio.plugins.dcm.DicomImageReader.colorSpaceOfFrame(DicomImageReader.java:591)
	at org.dcm4che3.imageio.plugins.dcm.DicomImageReader.applyColorTransformations(DicomImageReader.java:501)
	at org.dcm4che3.imageio.plugins.dcm.DicomImageReader.read(DicomImageReader.java:469)
```

**Solution:**
When `IllegalArgumentExpcetion` raises during getting ICCProfile,log warning and act as there is no ICCProfile specified.

### DIFF
--- a/dcm4che-image/src/main/java/org/dcm4che3/image/ICCProfile.java
+++ b/dcm4che-image/src/main/java/org/dcm4che3/image/ICCProfile.java
@@ -44,6 +44,8 @@ package org.dcm4che3.image;
 import org.dcm4che3.data.Attributes;
 import org.dcm4che3.data.Sequence;
 import org.dcm4che3.data.Tag;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.awt.color.ColorSpace;
 import java.awt.color.ICC_ColorSpace;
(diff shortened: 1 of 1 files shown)


ivosek:
@gunterze please can you have a look at this and consider if this can be merged?

if this is not acceptable solution for you, custom Exception would solve this for us as well, we can catch this specific Exception and clear value of tag ICCProfile, when this occurs

PR #1570, opened 2026-04-01
Add support for TCP Proxy protocol

shoraitek:
Changes allow for the parsing of Proxy Protocol from HAProxy allowing for blacklisting of source client IP's based on the passed on header.
If no proxy protocol header is present then system will fall back to standard communication, this means with --proxy set the system can handle non-proxy traffic as well.

usage:
./storescp --proxy -b TESTVNA:11111

log entry if active
09:08:42.531 INFO  - PROXY protocol enabled via command line

Testing 
Haproxy server IP : 10.140.160.201
Test client machine IP : 10.140.160.244

log entries for a normal DICOM echo traffic no proxy header
09:09:22.381 INFO  - read Proxy protocol
09:09:22.383 INFO  - Proxy read : /10.140.160.201:57818
09:09:22.384 INFO  - Accept connection from ProxySocket[proxied=/**10.140.160.201**:57818, receiving=/10.140.160.201:57818, delegate=Socket[addr=/10.140.160.201,port=57818,localport=11111]]

log for entry received from Haproxy with "send-proxy" enabled
09:10:16.718 INFO  - read Proxy protocol
09:10:16.719 INFO  - Proxy read : /10.140.160.201:46188
09:10:16.719 INFO  - Accept connection from ProxySocket[proxied=/**10.140.160.244**:61628, receiving=/10.140.160.201:46188, delegate=Socket[addr=/10.140.160.201,port=46188,localport=11111]]

log for entry received from Haproxy with "send-proxy-v2" enabled
09:11:16.112 INFO  - read Proxy protocol
09:11:16.114 INFO  - Proxy read : /10.140.160.201:55082
09:11:16.114 INFO  - Accept connection from ProxySocket[proxied=/**10.140.160.244**:61669, receiving=/10.140.160.201:55082, delegate=Socket[addr=/10.140.160.201,port=55082,localport=11111]]

this has been successfully tested with blacklisting and works well in blocking client IP


shoraitek:
Forgot to add the --help output for storeSCP

{3D23945A-A538-4E68-82A8-B56AF9453E33}

nhermab:
```/**
 * A {@link TCPProtocolHandler} that transparently handles the PROXY Protocol
 * (v1 and v2) on incoming TCP connections before delegating to the standard
 * dcm4che3 DICOM protocol handler.
 *
 * <h3>Why this class lives in {@code org.dcm4che3.net}</h3>
 * <p>{@link TCPProtocolHandler} is package-private in dcm4che3, so any
 * implementation must reside in the same package.</p>
 *
 * <h3>How it works</h3>
 * <ol>
 *   <li>Registered via {@link Connection#registerTCPProtocolHandler} for
 *       {@link Connection.Protocol#DICOM}, replacing the default
 *       {@code DicomProtocolHandler}.</li>
 *   <li>For each accepted connection, calls {@link ProxyProtocol#parse(Socket)}
 *       which reads (and strips) the PROXY header — or pushes the bytes back if
 *       no header is present.</li>
 *   <li>Wraps the result in a {@link ProxySocket} so that
 *       {@code socket.getInetAddress()} / {@code getRemoteSocketAddress()} return
 *       the <em>original client</em> IP rather than the load-balancer address.</li>
 *   <li>Delegates to the original {@code DicomProtocolHandler} which calls
 *       {@code Association.accept(conn, socket)} — the association now sees
 *       the real client IP.</li>
 * </ol>
 *
 * <h3>Usage</h3>
 * <pre>
 *   ProxyAwareConnection.enable();   // call before device.bindConnections()
 *   ProxyAwareConnection.disable();  // restore original handler
 * </pre>
 *
 * <h3>TLS note</h3>
 * <p>This handles plain-text TCP connections.  When deployed behind HAProxy
 * with PROXY protocol, TLS is typically terminated at the load-balancer level
 * before the PROXY header is prepended.</p>
 */
public class ProxyAwareConnection implements TCPProtocolHandler {

    private static final Logger LOG = LoggerFactory.getLogger(ProxyAwareConnection.class);

    private final TCPProtocolHandler delegate;

    private ProxyAwareConnection(TCPProtocolHandler delegate) {
        this.delegate = delegate;
    }

    /**
     * Enable PROXY protocol support globally for all DICOM connections.
     * Must be called <em>before</em> {@code device.bindConnections()}.
     *
     * <p>Replaces the default {@code DicomProtocolHandler} with a wrapper that
     * parses PROXY v1/v2 headers. If no header is present the connection is
     * handled normally (bytes are pushed back).</p>
     */
    public static void enable() {
        // Wrap the default DicomProtocolHandler with our proxy-aware handler.
        // registerTCPProtocolHandler returns the previous handler for this protocol.
        ProxyAwareConnection wrapper = new ProxyAwareConnection(DicomProtocolHandler.INSTANCE);
        TCPProtocolHandler previous = Connection.registerTCPProtocolHandler(
                Connection.Protocol.DICOM, wrapper);

        if (previous instanceof ProxyAwareConnection) {
            LOG.info("ProxyAwareConnection: was already enabled — re-registered");
        } else {
            LOG.info("ProxyAwareConnection: PROXY Protocol support ENABLED for DICOM connections (replaced handler: {})",
                    previous != null ? previous.getClass().getName() : "null");
        }
    }

    /**
     * Disable PROXY protocol support, restoring the original handler.
     */
    public static void disable() {
        Connection.registerTCPProtocolHandler(
                Connection.Protocol.DICOM, DicomProtocolHandler.INSTANCE);
        LOG.info("ProxyAwareConnection: PROXY Protocol support DISABLED");
    }

    @Override
    public void onAccept(Connection conn, Socket s) throws IOException {
        LOG.info("ProxyAwareConnection.onAccept: incoming socket from {} (raw remote={})",
                s.getRemoteSocketAddress(), s.getInetAddress());
        try {
            ProxyProtocol.Info info = ProxyProtocol.parse(s);

            LOG.info("ProxyAwareConnection: PROXY parse result: proxied={}, receiving={}, isProxied={}",
                    info.proxiedRemote, info.receivingRemote, info.isProxied());

            ProxySocket proxySocket = new ProxySocket(
                    s,
                    info.proxiedRemote,
                    info.receivingRemote,
                    info.pbInputStream);

            LOG.info("ProxyAwareConnection: handing off to DICOM handler — socket.getInetAddress()={}, socket.getRemoteSocketAddress()={}",
                    proxySocket.getInetAddress(), proxySocket.getRemoteSocketAddress());

            // Delegate to the original DICOM handler (Association.accept) with
            // the wrapped socket — getInetAddress() now returns the real client IP.
            delegate.onAccept(conn, proxySocket);

        } catch (IOException e) {
            LOG.error("ProxyAwareConnection: error parsing PROXY header from {}: {}",
                    s.getRemoteSocketAddress(), e.getMessage(), e);
            try { s.close(); } catch (IOException ignore) { /* best effort */ }
        }
    }
}
```

This is how i solved it.

PR #1589, opened 2026-06-08
Unprotected volatile access

sean-g-rae:
What: Access to a volatile variable (count) without concurrency protection
Why: Concurrent threads could have race conditions reading and/or writing values to the volatile variable.


sean-g-rae:
@gunterze Do I need to submit this to the dcm4che developers or is opening this PR adequate?  I also have some other PRs that add tests that are in the same boat.  Thanks!

